PR #83, opened 2015-05-30
Safeguard for regions without reads

bgruening:
During my testing I found one case where some regions do not have reads. This results in an empty SAM file and this crashes during SAM->BAM conversion. This PR should fix it.


bgruening:
@fidelram this fix is needed to pass all Galaxy tests. A point release would be nice at some point. Thanks.

bgruening:
Thanks!

PR #126, opened 2015-12-08
Devon changes

dpryan79:
Fixes all options from `computeMatrix`, `heatmapper`, and `profiler`. `--outFileNameData` removed from all tools, since it did nothing.

For computeMatrix, the following changes were made:
 * --missingDataAsZero changed such that it works
 * --maxThreshold/minThreshold fixed so they work
 * -q now suppresses more annoying messages
 * Documentation changes for --startLabel/--endLabel/--sortRegions/--referencePoint

For heatmapper, the following changes were made:
 * --perGroup actually does something now.
 * Documentation changes for --colorNumber

For profiler, the following changes were made:
 * --plotType overlapped_lines fixed
 * --plotType fixed to allow --perGroup
 * --xAxisLabel removed as an option, since there isn't one.
 * --colors Is now fully used.
 * --yMin with no --yMax no longer flips the axes.

The documentation changes are minimal at the moment, I'll do more after merging.


dpryan79:
Oh, multiple BED files are now supported.

dpryan79:
@bgruening: Note the loss of the `--outFileNameData` option for all tools and `--xAxisLabel` for `profiler`. These did nothing previously anyway.

bin/heatmapper
```diff
@@ -119,15 +119,26 @@ def prepare_layout(hm_matrix, heatmapsize, showSummaryPlot, showColorbar):
 
     numcols = hm_matrix.get_num_samples()
     numrows = hm_matrix.get_num_groups()
+    if perGroup:
+        temp = numcols
```
bgruening: a,b = b,a ? It's Python! http://xkcd.com/353/

bin/heatmapper
```diff
@@ -264,19 +278,26 @@ def plotMatrix(hm, outFileName,
...
+                    sample_idx = sample + 2 # plot + spacer
+                else:
+                    group += 2 # plot + spacer
                 first_group = 1
                                     
-            ax = fig.add_subplot(grids[group, sample])
-            if group == first_group and not showSummaryPlot:
+            if perGroup:
+                ax = fig.add_subplot(grids[sample_idx, group])
+            else :
```
bgruening: Style ... _:

bin/heatmapper
```diff
@@ -511,8 +555,8 @@ def main(args):
     if args.outFileNameMatrix:
         hm.saveMatrixValues(args.outFileNameMatrix)
 
-    if args.outFileNameData:
-        hm.saveTabulatedValues(args.outFileNameData)
+    #if args.outFileNameData:
```
bgruening: debug?

bin/heatmapper
dpryan79: I switch between enough languages that this is the only way I can maintain sanity :p

bin/heatmapper
dpryan79: The writer function was never written. This should be on the to do list for the next release, though.

bin/profiler
```diff
@@ -140,6 +144,22 @@ def plotProfile(hm, outFileName,
...
+        for plot in range(numplots):
+            col = plot % plots_per_row
+            row = int(plot / plots_per_row)
+            for data_idx in range(numlines):
+                if perGroup:
+                    row, col = plot, data_idx
+                else:
+                    row, col = data_idx, plot
+                sub_matrix = hm.matrix.get_matrix(row, col)
+                curMax = np.amax(np.__getattribute__(averagetype)(sub_matrix['matrix'], axis=0))
```
bgruening: This looks bad, is this API stable? ;)
I don't have it in my random install ...

bin/profiler
```diff
@@ -269,8 +301,8 @@ def main(args):
     if args.samplesLabel and len(args.samplesLabel):
         hm.matrix.set_sample_labels(args.samplesLabel)
 
-    if args.outFileNameData:
-        hm.saveTabulatedValues(args.outFileNameData)
+    #if args.outFileNameData:
```
bgruening: debug?

deeptools/computeMatrix.py
```diff
@@ -0,0 +1,418 @@
+#!/usr/bin/env python
+#-*- coding: utf-8 -*-
+
+import argparse
+import sys
+import os
+import os.path
```
bgruening: redundant ...

deeptools/computeMatrix.py
```diff
@@ -0,0 +1,418 @@
...
+To learn more about the specific parameters type:
+
+%(prog)s reference-point --help or
+%(prog)s scale-regions --help
+
+""",
+            epilog='An example usage is:\n  %(prog)s reference-point -S '
+            '<biwig file> -R <bed file> -b 1000\n \n')
+
+    parser.add_argument('--version', action='version',
```
bgruening: Shouldn't this be part of the main and imported here.

deeptools/computeMatrix.py
```diff
@@ -0,0 +1,418 @@
...
+    else:
+        args.verbose = False
+
+    if args.command == 'scale-regions':
+        args.nanAfterEnd = False
+        args.referencePoint = None
+    elif args.command == 'reference-point':
+        if args.beforeRegionStartLength == 0 and \
+                args.afterRegionStartLength == 0:
+            sys.stderr.write("\nUpstrean and downstream regions are both "
```
bgruening: Wouldn't `sys.exit("\nUpstrean and downstream regions are both ")` do the trick as well?

bin/profiler
dpryan79: Good question, I only new that it worked because Fidel used it elsewhere in the code. I agree that it should be replaced with something more standard.

deeptools/computeMatrix.py
```diff
@@ -0,0 +1,418 @@
...
+    hm = heatmapper.heatmapper()
+
+    scores_file_list = [x.name for x in args.scoreFileName]
+    hm.computeMatrix(scores_file_list, bed_file, parameters, verbose=args.verbose)
+    if args.sortRegions != 'no':
+        hm.matrix.sort_groups(sort_using=args.sortUsing, sort_method=args.sortRegions)
+
+    hm.saveMatrix(args.outFileName)
+    bed_file.close()
+    #os.remove(bed_file.name)
```
bgruening: debug?

deeptools/heatmapper.py
```diff
@@ -763,21 +781,18 @@ def getRegionsAndGroups(regions_file, onlyMultiplesOf=1,
...
-                        while True:
-                            i += 1
-                            newlabel = label + "_r" + str(i)
-                            if newlabel not in group_labels:
-                                break
-
-                    group_labels.append(label)
-                    group_boundaries.append(includedintervals)
+                group_idx += 1
+                label = ginterval.line[1:].strip()
```
bgruening: rstrip()? strip() before []?

deeptools/heatmapper.py
```diff
@@ -804,15 +819,28 @@ def getRegionsAndGroups(regions_file, onlyMultiplesOf=1,
...
 
+            ginterval.group_idx = group_idx
             regions.append(ginterval)
             includedintervals += 1
 
         # in case we reach the end of the file
         # without encountering a hash,
         # a default name is given to regions
-        if len(regions) > group_boundaries[-1]:
+        if len(group_labels) == 0:
```
bgruening: if len(group_labels):

bin/heatmapper
bgruening: Fair enough ... skip the other ones ;)

bin/heatmapper
fidelram: The function was not really working, thus I commented it out long time ago. Furthermore, for the users that want the raw data, they can always unzip the matrix that is passed from computeMatrix to heatmapper.

bin/profiler
fidelram: What is the problem here?

dpryan79: The last line, in particular `np.__getatrribute__(averagetype)(...)`.

deeptools/computeMatrix.py
fidelram: Here yes, that would be equivalent.

deeptools/computeMatrix.py
fidelram: I just fixed that a while ago while addressing another issue. Thanks for pointing out.

deeptools/heatmapper.py
fidelram: This is ok. The label could have white spaces either left or right, so strip() is the right choice. It is line[1:], because at position zero is the '#' that labels the groups.

bgruening: Ok, sorry for the noise!

deeptools/heatmapper.py
fidelram: Not in this case, we want to test that the group_labels is empty. But I just checked and the pythonic way is `if not group_labels` I will change that.

bin/profiler
fidelram: I like that way of calling np.sum, np.mean, np.median etc. with one line. Do you know a better way to do it?

deeptools/computeMatrix.py
fidelram: I do not clearly understand... the background is, I tried to keep each tool with its own documentation, except when the documentation is shared by other tools in which case it is located in deeptools/parserCommon.py

deeptools/computeMatrix.py
fidelram: I removed a bunch of those cases as part of my PEP8 clean up. Taking about PEP8, I still have a lot of camelCase variables that are harder to change.

deeptools/computeMatrix.py
bgruening: Yeah, was just wondering if this should be moved into parserCommon.

deeptools/computeMatrix.py
bgruening: Ah, nice. I also was working on this. Will stop now. Let me know if you need help here.

fidelram: I am done for today.

PR #127, opened 2015-12-09
Fix heatmapper tests

fidelram:
Code clean-up. Doctests were moved to unit tests and the tests were updated. Some PEP8 renamings were also done in the process.


deeptools/plotHeatmap.py
```diff
@@ -186,7 +186,7 @@ def plotMatrix(hm, outFileName,
 
     if zMax is None:
         if matrix_flatten is None:
-            matrix_flatten = flattenMatrix(hm.matrix.matrix)
+            matrix_flatten = hm.matrix.matrix.flatten()
```
bgruening: This looks suspicious to me.

fidelram: the idea is to flatten the matrix only one. thus, be default is set to None. I replaced this by `if not matrix_flatten`

dpryan79: I think the question was whether you instead meant `matrix_flatten = hm.matrix.flatten()`.

bgruening: Yes, what @dpryan79 was saying. Sorry, I will be more specific next time.

fidelram: I see. I changed that in the new PR.

Issue #110, opened 2015-10-29
add region labels to the heatmap


friedue:
Currently, the information about which region is plotted is only shown in the profiler plots. I.e., if one chooses to only show the heatmap plus colorbar, this information is lost/not visible. 
Perhaps adding an extra bar/line on top of the heatmap (not making it part of the heatmap itself) might help solve this issue?

dpryan79:
I think this is fixed in the `devon_changes` branch, which will hopefully get merged today.

fidelram:
It is now merged. Friederike, do you still have the problem?

On Tue, Dec 8, 2015 at 11:49 AM, Devon Ryan <notifications@github.com>
wrote:

> I think this is fixed in the devon_changes branch, which will hopefully
> get merged today.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/fidelram/deepTools/issues/110#issuecomment-162846920>.
>



-- 

Fidel Ramirez

PR #130, opened 2015-12-09
Added test dependent on matplotlib version. Added labels and colorbar to plotProfile(#110)

fidelram:
Now plotProfile can generate images like the following:

large-h

The user can change the scale, which is equal to all plots. The color map can not be changed. 

This PR should solve #110


deeptools/plotProfile.py
```diff
@@ -134,53 +134,145 @@ def __init__(self, hm, out_file_name,
...
 
         # convert cm values to inches
-        plot_height_inches = rows * float(self.plot_height) / 2.54
-        fig_width = cols * float(self.plot_width) / 2.54
-        fig = plt.figure(figsize=(fig_width, plot_height_inches))
+        plot_height_inches = rows * self.cm2inch(self.plot_height)[0]
+        self.fig = plt.figure(figsize=self.cm2inch(cols * self.plot_width, rows * self.plot_height))
+        self.fig.suptitle(self.plot_title, y=(1 - (0.06 / plot_height_inches)))
+
+    @staticmethod
```
bgruening: Why do we need a staticmethod here?

fidelram: because no 'self' variable is used.

PR #131, opened 2015-12-09
add debug message

fidelram:
fix all tests that were not passing in travis. Added doctests that travis was not running.


bgruening:
Awesome! @fidelram together with https://github.com/fidelram/deepTools/pull/132 we should have running tests :)

PR #132, opened 2015-12-09
WIP: flake8 testing

bgruening:


bgruening:
@fidelram @dpryan79 can I request a careful review? 
This PR will fix all pep8 errors, but
* E501 - 79 character limit
* F403 - from foo import *

E501 is ok nowadays, but F403 needs to be fixed at some point. But I feel this can break some things, so we need to be careful.
Especially https://github.com/fidelram/deepTools/commit/7cf1bec32e4fdc8b526390449d7ec26057c12d4a contains some important fixes. Actually, I think some part where not working before. We should enforce pep8 testing to catch those errors from now on. 
Please review carefully before merging. If this PR has a too big impact to the code base at this point let me know. Also feel free to cherry pick commits if needed.

bgruening:
Green! It's green!
@fidelram thanks for the rebase!

deeptools/heatmapper_utilities.py
```diff
@@ -15,9 +15,9 @@ def plot_single(ax, ma, average_type, color, label, plot_type='simple'):
         ax.fill_between(x, sumry, facecolor=color, alpha=0.6)
 
     elif plot_type == 'se':  # standard error
-        std = np.std(ma, axis=0)/np.sqrt(ma.shape[0])
+        std = np.std(ma, axis=0) / np.sqrt(ma.shape[0])
         alpha = 0.2
-        if type(color) == type((0, 0)):  # check if color is tuple
+        if isinstance(color, type((0, 0))):  # check if color is tuple
             # add the alpha channed to the color tuple
             f_color = [c for c in color[:-1]] + [alpha]
```
fidelram: Here there could be a potential problem because the check `if isinstance(color, type((0, 0))):` would return true regardless of the length of `color`. However, whenever the color is a tuple I suppose it is always a proper color tuple with three elements. In other words, the test, I think, wants to check if `color` is tuple or `str`. I will check this

PR #135, opened 2015-12-10
Remove all references to and use of bigWigInfo

dpryan79:
We really don't need `bigWigInfo` anymore. In fact, it was breaking things for Vivek since kent's tools aren't as reliable as pyBigWig at remote file access.

### DIFF
--- a/config/deepTools.cfg
+++ b/config/deepTools.cfg
@@ -2,8 +2,6 @@
 sort: sort
 samtools: samtools
 bedgraph_to_bigwig: bedGraphToBigWig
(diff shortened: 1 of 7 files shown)


bgruening:
LGTM. Let me know when you remove the WIP label.

dpryan79:
I removed the WIP label, I suspect one can use set intersections and some iterators, but while that's more pythonic I'm not sure it's more efficient. So we can stick with the old code.

bgruening:
Awesome! Thanks!

PR #142, opened 2015-12-12
remove fragmentLength

bgruening:
Is this correct @fidelram?

This tool does not work with this option and as far as I can see it can be removed.


dpryan79:
That should have been subsumed by the `--extendReads` parameter, or so I thought.

bgruening:
Feel free to merge :)

fidelram:
Thanks for taking care of this.

PR #143, opened 2015-12-12
bwh.chroms() return a dict

bgruening:
@dpryan79 this one is for you :)

bwh.chroms() return a dict and we need to iterate over keys and values.


dpryan79:
It disturbs me that this wasn't caught by a test, but you're absolutely correct.

bgruening:
Soon I have all tools at least under Galaxy tests ... but these are not meaningful at all.
Thanks for merging.

dpryan79:
We're at least putting together an internal battery of tests, so that will hopefully cover this.

Issue #144, opened 2015-12-12
--missingDataAsZero is not a boolean


bgruening:
@fidelram @dpryan79 do we want to take the opportunity and make a boolean out of it? Maybe `--disableMissingDataAsZero`? It's long and ugly, but the `yes`/`no` is not intuitive either.

```
    optional.add_argument('--missingDataAsZero',
                          default="yes",
                          choices=["yes", "no"],
                          help='Default is "yes". This parameter determines '
                          'if missing data should be treated as zeros. '
                          'If set to "no", missing data will be ignored '
                          'and not included in the output file. Missing '
                          'data is defined as those regions for which '
                          'both BAM files have 0 reads.')
```

bgruening:
Especially because in computeMatrix we have this: https://github.com/fidelram/deepTools/blob/release-1.6/deeptools/computeMatrix.py#L248

vivekbhr:
That's right I would vote for the boolean option..

vivekbhr:
@bgruening @fidelram @dpryan79 the option can be named : --keepMissingData , and we can edit the help to say what it means

bgruening:
@vivekbhr I like `--keepNAs`.
This option is in many places, even referenced in other parameters so we need to grep through the code once we agreed on a way forward.

fidelram:
we should make this boolean!

PR #145, opened 2015-12-13
new Galaxy wrappers for version 2.0

bgruening:
depends on https://github.com/fidelram/deepTools/issues/144


bgruening:
@fidelram, @dpryan79 here is a package with the latest 1.6 branch: https://testtoolshed.g2.bx.psu.edu/view/iuc/package_python_2_7_deeptools_2_0
If this PR gets merged we just need to rebuild the container and get always the latest 1.6 branch with all bugfixes. As soon as we tag a 2.0 release we simply change the URL in the tool_dependencies.xml file and we should be ready.

dpryan79:
Can you remove emf from the list of image formats? Matplotlib doesn't actually support it in Galaxy (not that anyone uses it anyway).

bgruening:
Done!

dpryan79:
LGTM!

bgruening:
Great, I will wait until the blocking issue is fixed, this will streamline the handling of tools.

fidelram:
@bgruening #144 has been addressed, now --keepNAs is in place. Also the --MNAse parameter of bamCoverage is working.

bgruening:
Out of WIP, at least as a first version.

dpryan79:
Cool! We have a local instance of this now, so I'll look at the live versions tomorrow and start testing :+1:

galaxy/wrapper/deepTools_macros.xml
```diff
@@ -308,23 +380,25 @@ This tool is developed by the `Bioinformatics and Deep-Sequencing Unit`_ at the
...
-            <option value="emf">emf</option>
         </param>
     </xml>
 
-    <xml name="missingDataAsZero">
-                <param name="missingDataAsZero" type="boolean" truevalue="yes" falsevalue="no" checked="True"
-                    label ="Treat missing data as zero"
-                    help  ="This parameter determines if missing data should be treated as zeros. If unchecked, missing data will be ignored and not included in the output file. Missing data is defined as those regions for which both BAM files have 0 reads." />
+    <xml name="keepNAs">
+        <param argument="--keepNAs" type="boolean" truevalue="--keepNAs" falsevalue="" checked="False"
```
fidelram: @bgruening we renamed this to be --skipNonCoveredRegions (or simply --skipNAs). 

The help says:

 'This parameter determines if non covered regions regions without overlapping reads) in a bam file should 
  be skipped. The default is to treat those regions as having a value of zero.  The decision to skip non 
 covered regions depends on the interpretation of the data. Non covered regions may represent for 
 example repetitive regions that want to be skipped.

PR #146, opened 2015-12-14
changed parameter --missingDataAsZero to a boolean to address #144

fidelram:
Some extra function renamings were added. I didn't change the name of the variable as this is in the documentation as well and would be an overhead at the moment.


fidelram:
I am now renaming to --keepNAs

fidelram:
@bgruening I think is safe to merge now. I am working on a fix for the failing tests on a separate branch. Noticed that I also change the parameter name from computeMatrix and for bigwigCompare.

Issue #124, opened 2015-12-08
Don't label lines for some plot types


dpryan79:
Some plot types (e.g., overlapped_lines) don't support line labels. Trying to set them causes a needless warning to be printed to the screen. This could just be prevented with a small `if ...` at a few places.

Issue #107, opened 2015-10-19
profiler v1.6 issues


mpi-pebert:
Hi,
I have a couple of remarks concerning the profiler tool (checked out from the dev branch):

1) title: an added title overlaps with the plot(s), there needs to be more white space at the top
2) colors: specifying colors as hex is a bit "tricky"; if one does not put the "#" in front, the string is not recognized as hex color (ok), however, with the "#' in front, the user has to remember to double quote the string if the profiler call is to be saved in a shell script. One could think of making the handling of hex color strings more user-friendly
2.1) when making a profiler call for N marks, specifying N colors and setting
```python
--plotFileFormat png --averageType median --plotType overlapped_lines --perGroup
```
the output looks more or less always like this:

![image](https://cloud.githubusercontent.com/assets/5285064/10573699/bbdd29a2-7651-11e5-87c9-6cda02b2c589.png)

Yes, it is overlapping the lines, but not recognizing/using the colors. The black background is default, then?

3) the help for profiler currently mentions the option
*this option only works if --onePlotPerGroup is set* in the description of *--plotType*; I presume this is now *--perGroup*?

4) to me, it is not really clear how this grouping is supposed to work. It sounds like I could somehow specify N datasets from sample A and M datasets from sample B and then plot two profiles next to each other, one for sample A with all N data points/lines superimposed and one for sample B with all M data points/lines, respectively. But I do not understand how to do that; maybe a glossary concerning the use of the words *sample* and *group* (and potentially others) on the Wiki would be helpful?

That's all for now, looking forward to the 2.0 release!

fidelram:
How many bigwig files did you use? Did you have some groups?

mpi-pebert:
I used 8 bigWig files in that case; different marks from the same sample, so in my understanding, all belong to the same group. However, as I hinted at under point 4 above, the "what" and "how" of defining groups is not entirely clear to me ;-)

dpryan79:
The automatic closing of issues is currently broken, so I'll close this manually.

brianpenghe:
Your Remark 2 solved my puzzle. Thank you so much!

PR #150, opened 2015-12-15
replaced the plotType *overlapped_lines* to address #107 & #124

fidelram:
Now I am using matplotlib hexbin which I think results in better images. Not fully tested. The result now looks like this.

overlapped_lines


dpryan79:
N.B., I reverted this since it got rid of a previously tested and working option in the middle of a feature freeze. Don't delete this branch, since it can be merged back in after the Galaxy stuff is dealt with.

PR #151, opened 2015-12-15
Modification of code structure to allow sphinx documentation on correctGCbias and computeGCbias

steffenheyne:
can you check if I change the arg parsing correctly? It works locally


fidelram:
Test for correctGCbias fails because samtools is not installed in travis.

bgruening:
Added samtools to travis.

PR #156, opened 2015-12-16
changed file structure of bamPEFragmentSize and added documentation

fidelram:


dpryan79:
LGTM

deeptools/bamPEFragmentSize.py
```diff
@@ -0,0 +1,91 @@
...
+    print "Min.: {}\n1st Qu.: {}\nMean: {}\nMedian: {}\n" \
+          "3rd Qu.: {}\nMax.: {}\nStd: {}".format(fragment_len_dict['min'],
+                                                  fragment_len_dict['qtile25'],
+                                                  fragment_len_dict['mean'],
+                                                  fragment_len_dict['median'],
+                                                  fragment_len_dict['qtile75'],
+                                                  fragment_len_dict['max'],
+                                                  fragment_len_dict['std'])
+
+    print "\nRead lengths:"
```
bgruening: We should use logging here, right?

dpryan79: That's why the unix gods created `>` :)

bgruening: ;) NoNo!

fidelram: Why would logging improve in this case? Do you need the output in certain way for galaxy?

PR #163, opened 2015-12-18
documentation update

fidelram:


dpryan79:
N.B., I'll obviously need to proof read these, so I'll do the merging once this is out of WIP if that's OK with everyone.

PR #164, opened 2015-12-18
Updated help text on extendReads

kilpert:


dpryan79:
What happens if 0 is used? Thomas mentioned getting an error message. Perhaps that should be noted too?

fidelram:
The default behaviour when the extension length is smaller than the read length  is to print a warning message and do no any extension.

PR #191, opened 2015-12-22
Support multiple bigWig/BED files in computeMatrix. Also, remove the …

### LINKED ISSUE #179: computeMatrix: advanced output settings
one of the "advanced output" is now very purpose of the tool (matrix of values).
Not sure if the other outputs (profile data and regions after skipping) are still reported.

### LINKED ISSUE #186: computeMatrix should take multiple score files and produce a single output
`computeMatrix` now takes `-b file1.bw file2.bw ...` and produces a single matrix from that. This isn't currently supported.

dpryan79:
…saveData option from all tools, since it doesn't exist

Fixes #186
Fixes #179

### DIFF
--- a/galaxy/wrapper/computeMatrix.xml
+++ b/galaxy/wrapper/computeMatrix.xml
@@ -9,28 +9,22 @@
 <![CDATA[
         #import tempfile
 
-        #for $rf in $regionsFiles:
-            cat "$rf.regionsFile" >> ./temp_input_path;
-            #if str($rf.label.value).strip():
-                echo "\#$rf.label.value" >> ./temp_input_path;
-            #else:
-                echo "\#$rf.regionsFile.name" >> ./temp_input_path;
-            #end if
-        #end for
-
         @BINARY@
 
             $mode.mode_select
-            --regionsFileName ./temp_input_path
-            --scoreFileName '$scoreFile'
+            --regionsFileName
+            #for $rf in $regionsFiles:
+                '$rf.regionsFile' 
+            #end for
+            --scoreFileName
+            #for $bw in $scoreFiles:
+                '$bw.scoreFile' 
+            #end for
             --outFileName '$outFileName'
 
             @THREADS@
 
             #if $output.showOutputSettings == "yes"
-                #if $output.saveData:
-                    --outFileNameData '$outFileNameData' 
-                #end if
                 #if $output.saveMatrix:
                 --outFileNameMatrix '$outFileNameMatrix'
                 #end if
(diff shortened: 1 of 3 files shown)


galaxy/wrapper/computeMatrix.xml
```diff
@@ -85,12 +79,14 @@
                 help="Label to use in the output."/>
         </repeat>
 
-        <param name="scoreFile" format="bigwig" type="data"
```
bgruening: Alternative. Please note the multiple=True.
Moreover, why does the `-scoreFile` only have one dash? Needs a fix.
```
<param argument="-scoreFile" format="bigwig" type="data" multiple="True"
           label="Score file"
           help="Should be a bigWig file (containing a score, usually covering
           the whole genome). You can generate a bigWig file either from a
           bedGraph or WIG file using UCSC tools or from a BAM file using the
           deepTool bamCoverage."/>
```

dpryan79: --scoreFileName or -S

dpryan79: I have no clue how multiple="True" is handled in Galaxy, it's not documented. We need the quotes to be handled correctly and my way guarantees that.

bgruening: Do you mind this change? Otherwise I can implement it. You don't want a repeat block if you insert >5 files also this enables a collection as input.

bgruening: We are using it already here: https://github.com/fidelram/deepTools/blob/release-1.6/galaxy/wrapper/deepTools_macros.xml#L292

dpryan79: It's fine as long as it works correctly. I won't personally implement things that I can't find documented anywhere in the wiki.

galaxy/wrapper/computeMatrix.xml
```diff
@@ -9,28 +9,22 @@
...
 
             $mode.mode_select
-            --regionsFileName ./temp_input_path
-            --scoreFileName '$scoreFile'
+            --regionsFileName
+            #for $rf in $regionsFiles:
+                '$rf.regionsFile' 
+            #end for
+            --scoreFileName
+            #for $bw in $scoreFiles:
```
bgruening: The corresponding change here would be:

```
#for $bw in $scoreFile:
    '$bw'
#end for
```

galaxy/wrapper/computeMatrix.xml
bgruening: https://wiki.galaxyproject.org/Admin/Tools/ToolConfigSyntax#A.3Cparam.3E_tag_set under type="data" you can find documentation.

dpryan79:
@bgruening: Anything else that needs to be changed?

bgruening:
Yes thanks.

PR #206, opened 2016-01-05
Fix #205 in the next release

### LINKED ISSUE #205: 1.5.13 is too big
The 1.5.13 release is over 40MB big. It would be nice to not include the webpage into the release tarball. Also the last stable release is not on pypi.

dpryan79:
This drastically shrinks the size of the release tarball that can get uploaded to pypi (from ~17MB to <1MB).

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,8 +2,5 @@ include *.txt
 include README.md
 include deeptools/config/deeptools.cfg
 include docs/manual/*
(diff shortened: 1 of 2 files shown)


dpryan79:
I should note that the galaxy stuff was never getting uploaded to pypi (and doesn't need to be).

bgruening:
:+1:

PR #207, opened 2016-01-05
Fix198

### LINKED ISSUE #198: Support remote BAM/bigWig files
This largely already works in the release-1.6 branch, but there's no reason that deepTools can't now support remote bigWig/BAM files.

dpryan79:
Ensure that remote file access works for everything. Fixes #198

### DIFF
--- a/bin/estimateScaleFactor
+++ b/bin/estimateScaleFactor
@@ -30,7 +30,6 @@ def parseArguments(args=None):
                         metavar='list of bam files',
                         help='List of indexed BAM files, space delineated',
                         nargs='+',
-                        type=fileType,
                         required=True)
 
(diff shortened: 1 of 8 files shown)


bgruening:
We used that in Galaxy many years ago.
The tests are failing @dpryan79.

dpryan79:
Yeah, I noticed, this is a WIP so I'll fix the tests at the end. It looks like bamHandler is importing things for other files.

dpryan79:
Remote files now work if one uses the WriterIntegration branch of pyBigWig, which I expect to merge into master and tag as version 0.2.0 within the next week. I don't know if we really need to make that version a requirement or not.

PR #212, opened 2016-01-07
Fix199

### LINKED ISSUE #199: Remove dependency on UCSCtools/samtools
Now that deepTools uses pyBigWig, there's no need for UCSCtools. While not yet completely documented, pyBigWig now supports bigWig file creation without intermediate files. Also, there's nothing that samtools does that we can't just do with pysam.

These two changes would considerably simplify the installation as well.

dpryan79:
This fixes #199 but will only work once pyBigWig 2.0 is released. My minimal testing suggests that this should work, but we should really test further before merging. Thus, this will be a WIP for a while.

### DIFF
--- a/deeptools/correctGCBias.py
+++ b/deeptools/correctGCBias.py
@@ -16,11 +16,8 @@ import argparse
 from scipy.stats import binom
 
 from deeptools.utilities import getGC_content, tbitToBamChrName
-from deeptools import config as cfg
 from deeptools import writeBedGraph, parserCommon, mapReduce
 from deeptools import utilities
-
-samtools = cfg.config.get('external_tools', 'samtools')
 
 
 def parse_arguments(args=None):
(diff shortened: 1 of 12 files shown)


bgruening:
Ok, for what it's worth ... all or nothing I would say. We depend on this new version, rerunning all our tests and doing the release. For the Galaxy Servers this feature is not essential, but for commandline folks this is nice to have. So if it does not break old stuff, let's merge it before releasing. Just my 5c.

deeptools/correctGCBias.py
```diff
@@ -617,17 +593,18 @@ def main(args=None):
...
-                                                " ".join(res))
+            print "concatenating (sorted) intermediate BAMs"
+            of = pysam.open(args.correctedFile.name, "wb", template=res[0])
+            for f in res:
+                for e in f.fetch(until_eof=True):
+                    of.write(e)
+            of.close()
 
-        run_shell_command(command)
+        print "indexing BAM"
```
bgruening: For 2.1 we should get rid of all print statements and use logging.

bgruening:
Nice cleanup! :+1:  from my side.

deeptools/correctGCBias.py
dpryan79: Is there really any benefit to using logging for stuff like this?

dpryan79:
We'll be testing this locally for the next few days to hopefully catch any introduced bugs.

PR #220, opened 2016-01-15
renaming of tools to better match their purpose

fidelram:
* bamCorrelate→ multiBamCoverage
 * bigwigCorrelate → multiBigwigSummary
 * bamFingerplot → plotFingerprint


bin/multiBamCoverage
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 #-*- coding: utf-8 -*-
 
-from deeptools.bamFingerprint import main
+from deeptools.multiBamCoverage import main
```
fidelram: Git went crazy with those renames. But they end files are correct any way.

bgruening:
Fidel please let this open, I will try to fix the Galaxy tools as well and push it into here, if you don't mind.

fidelram:
@bgruening Sure

fidelram:
@bgruening I am not done with the renaming in the documentation yet.

fidelram:
@bgruening I am done. All yours!

bgruening:
All Galaxy tests are failing because the resulting bw files are slightly different then before. This is probably unrelated to this PR, but is this expected?
ping @dpryan79

dpryan79:
Yes, that's expected. Did you add pyBigWig-0.2.3 or to Galaxy? I actually have PR ready for the IUC if not.

bgruening:
Not yet, installed via pip this branch.

dpryan79:
That should take care of the install then, unless you specified no dependencies. UCSC and samtools are no longer required, but a very recent version of pyBigWig is. That's available in pypi and should get installed automatically unless otherwise specified.

But yeah, the results should differ, since there will be differences in zoom levels and number of children per index node in the bigWig file.

bgruening:
Ok, this is good, will update the test-data. Thanks!

bgruening:
Tests are now all green in Galaxy. @fidelram I changes some things related to actual code so you might have a look at it.

fidelram:
Everything seems fine. I will merge now.

PR #222, opened 2016-01-16
One should not need to be in the source code directory for nosetests …

dpryan79:
…to work

@bgruening: This increases the tarball size to ~850KB. I see that as a reasonable trade-off, particularly given that meaningful nosetests will help with bioconda integration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,6 @@ install:
 # command to run tests
 script:
   - flake8 . --exclude=.venv,.build --ignore=E501,F403,E402
-  - nosetests --with-doctest deeptools/test `ls deeptools/*.py | grep -v correct`
+  - nosetests --with-doctest deeptools/test deeptools --ignore-files="correctGCBias"
 
 sudo: false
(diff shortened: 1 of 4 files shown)


bgruening:
Looks awesome. I'm just waiting for the pypi release to push the conda deepTools package!

dpryan79:
Yeah, I only thought about this since I was playing with a conda build and the `nosetests` that would work were not very meaningful.

bgruening:
nosetest are really nice and conda integration is awsome! :)
This is getting a very nice release!

.travis.yml
```diff
@@ -27,6 +27,6 @@ install:
 # command to run tests
 script:
   - flake8 . --exclude=.venv,.build --ignore=E501,F403,E402
-  - nosetests --with-doctest deeptools/test `ls deeptools/*.py | grep -v correct`
+  - nosetests --with-doctest deeptools/test deeptools --ignore-files="correctGCBias"
```
fidelram: I had to give explicitly `ls deeptools/*.py` because nosetests was not running all doctest by default when simply giving the directory. Does it work for you?

dpryan79: More or less, though `deeptools/test` should be `deeptools.test`. It runs 66 tests, at least. I'll figure out what the missing 3 are and add them manually.

PR #223, opened 2016-01-17
Remove args.bai

bgruening:
This was removed from the function a few days ago. This made same Galaxy tests fail.

### DIFF
--- a/deeptools/bamPEFragmentSize.py
+++ b/deeptools/bamPEFragmentSize.py
@@ -17,12 +17,6 @@ def parse_arguments():
                         help='BAM file to process',
                         metavar='bam-file')
 
-    parser.add_argument('-bai',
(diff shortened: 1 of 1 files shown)


deeptools/bamPEFragmentSize.py
```diff
@@ -17,12 +17,6 @@ def parse_arguments():
                         help='BAM file to process',
                         metavar='bam-file')
 
-    parser.add_argument('-bai',
-                        help='Index for the BAM file. Default is to '
-                        'consider the path of the BAM file adding '
-                        'the .bai suffix.',
-                        metavar='bam-file-index')
-
```
dpryan79: I thought I already removed all of these, good catch!

bgruening: Your welcome :)

PR #224, opened 2016-01-18
Redo nose fixes with the renamed files. This is simpler than merging :)

dpryan79:


fidelram:
I take that the reason for nose to skip some files was permissions. Is that right?

dpryan79:
Yeah, nose will skip any file that's executable.

PR #227, opened 2016-01-18
Fix smooth length

### LINKED ISSUE #226: --smoothLength is ignored
`bamCoverage` and `bamCompare` have a `--smoothLength` option that ends up getting ignored.

dpryan79:
Fix #226

### DIFF
--- a/deeptools/bamCompare.py
+++ b/deeptools/bamCompare.py
@@ -305,7 +305,7 @@ def main(args=None):
                                      verbose=args.verbose
                                      )
 
-    wr.run(FUNC, func_args, args.outFileName, format=args.outFileFormat, smooth_length=args.smoothLength)
+    wr.run(FUNC, func_args, args.outFileName, format=args.outFileFormat, smoothLength=args.smoothLength)
 
 if __name__ == "__main__":
     main()
--- a/deeptools/bamCoverage.py
+++ b/deeptools/bamCoverage.py
@@ -215,7 +215,7 @@ def main(args=None):
                                          )
 
     wr.run(writeBedGraph.scaleCoverage, func_args, args.outFileName,
-           format=args.outFileFormat, smooth_length=args.smoothLength)
+           format=args.outFileFormat, smoothLength=args.smoothLength)
 
 
 class CenterFragment(writeBedGraph.WriteBedGraph):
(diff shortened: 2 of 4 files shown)


deeptools/bamCompare.py
```diff
@@ -305,7 +305,7 @@ def main(args=None):
                                      verbose=args.verbose
                                      )
 
-    wr.run(FUNC, func_args, args.outFileName, format=args.outFileFormat, smooth_length=args.smoothLength)
+    wr.run(FUNC, func_args, args.outFileName, format=args.outFileFormat, smoothLength=args.smoothLength)
```
bgruening: I liked the former name more. Not sure if we have a style guideline.

dpryan79: I prefer camel case, but regardless `smoothLength` is the name of the argument as well and, therefore, requires fewer changes.

dpryan79:
There's still an issue where `--scaleFactorsMethod SES` fails on the test data, likely because the BAM files have all of the chromosomes in the header but only data for two of them. I'm not familiar enough with how SES works to handle that immediately, but if it's not fixed by tonight I'll just figure it out and fix it.

deeptools/bamCompare.py
fidelram: Here is a bit of history: first deeptools programs used camel case, then I learned about the python style guidelines (Devon's dreaded PEP8) that do not recommend to use the camel case. So after some point is stopped using camel case. You can tell new code from old code by this.  However, for the command line arguments I didn't make that change to keep some consistency.

fidelram:
The best fix is to have the proper headers in the bam files. SES samples the genome to estimate the background read coverage. To define the sampling sites it uses the bam headers to decide where to sample. When no data is available for a chromosome SES does not knows and just reports zeros. 

I don't know how frequent is the case that the bam headers have more chromosomes than the alignment in the bam except in tests. Thus, I don't think is worth to spend time fixing this for a test, rather than modifying the test.

dpryan79:
Err, those are the proper headers. I imagine that this sort of use-case is rather rare for deepTools, but in other types of *-seq splitting files by chromosome and processing each on separate machine is common (this is how large-scale variant calling is often done). In those cases no one bothers to modify the BAM header, there's no point.

Anyway, I'll merge this and think more about whether it's worth addressing the SES issue (and numberOfSamples) issue.

fidelram:
For deepTools the solution is to count the number of reads per chromosome and filter those chromosomes without reads. Calling idxstats from pysam can provide this information but this takes a while as this info is not readily available but needs to be computed.

PR #238, opened 2016-01-22
Remove the non-existant outFileCorMatrix option from the plotPCA code…

dpryan79:
…. Checking for the argument breaks the function


bgruening:
I wonder why the Galaxy tests have passed with this one :(

dpryan79:
Me too. I also wonder why our internal test was marked as having passed.

PR #247, opened 2016-01-27
use pyBigWig chrom names to translate between chr and non-chr chromosomes

fidelram:
Since pyBigWig now provides chromosome names, I changed the code to replace/remove 'chr' in case the BED file and the bigwig file chromosome names didn't match. Also, since pyBigWig also reports the chromosome length, I fixed a block caused when a region was extended beyond the chromosome size or when the chromosome name was not in the bigwig file (#247)


deeptools/heatmapper.py
```diff
@@ -483,48 +502,42 @@ def coverage_from_big_wig(bigwig, chrom, zones, binSize, avgType, nansAsZeros=Fa
...
-                                 "bigWig file.\nAn empty line will be "
-                                 "added you your heatmap.\nThe offending "
-                                 "chromosome name is "
-                                 "{}\n\n".format(chrom))
 
+        # adjust bw_array if it extends beyond chromosome limits
         if bw_array is not None:
-            if zones[0][0] < 0:
-                values_array = np.zeros(zones[-1][1] - zones[0][0])
-                values_array[:] = np.nan
```
fidelram: Removed lines because in line 501 they were already defined

Issue #231, opened 2016-01-19
Offer the option to export the data used to generate plots


friedue:
Most tools have the option already, but the ones currently missing it are:

* [ ] plotPCA
* [x] plotProfiles

fidelram:
plotProfiles can now save the data.

dpryan79:
Have a look at PR #289 and let me know if the text output suffices (it's easiest to just look at the changed file, since I have an example output in a Galaxy wrapper test).

dpryan79:
This is now fixed for the next release. Closing.

PR #252, opened 2016-01-28
allow saving of plotProfile data #231

fidelram:
Apart from saving the data, I also did some make up changes to the profile plots. For the plot type = 'overlapped lines' I added the labels that were missing. For the plot type = 'heatmap' I rotated the group labels.

### DIFF
--- a/deeptools/heatmapper.py
+++ b/deeptools/heatmapper.py
@@ -697,7 +697,6 @@ class heatmapper(object):
                 xtickslabel.append('{0:.1f}{1}'.format(float(a) / quotient, symbol))
 
             xticks = [(k / w)  for k in xticks_values]
-	print(xticks, xtickslabel)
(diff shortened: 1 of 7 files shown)


deeptools/heatmapper.py
```diff
@@ -647,33 +647,75 @@ def save_matrix(self, file_name):
...
+
+            xticks = [(k / w)  for k in xticks_values]
+        x_axis = np.arange(xticks[-1]) + 1
+        labs = []
+        for x_value in x_axis:
+            if x_value in xticks:
+             labs.append(xtickslabel[xticks.index(x_value)])
+            else:
+             labs.append("")
+
```
fidelram: All this code above is simply to get some labels for the x-axis of the output that match what is plotted. Otherwise is just bins ids.

Issue #248, opened 2016-01-27
validate correctGCBias produced BAM files


dpryan79:
This is basically a note to remind me to look sometime at the BAM files produced by correctGCBias. Since it's not playing around with the flags, they should pass htsjdk's rather absurd validator as long as the input file does as well. However, this may not always be the case and it's actually quite simple to just munge the flags such that they're valid (or at least produce a little warning that the BAM file is valid, but IGV/Picard/GATK/etc. will possibly bork). Cf. https://groups.google.com/forum/#!topic/deeptools/2gufhgs2YBQ

dpryan79:
It looks like correctGCBias is also adding invalid auxiliary tags (`CO:i` and `GC:f`) and misusing some legit ones (`CP:i`). This should be corrected at the same time.

dpryan79:
This should be fixed by #253

PR #253, opened 2016-01-28
Ensure that auxiliary tags maintain their type. Also, ensure that the…

dpryan79:
… added tags are legit. Note that this may not work in pysam<0.8.2


dpryan79:
This alone will likely fix #248

dpryan79:
N.B., see also https://github.com/pysam-developers/pysam/pull/214, which should get rid of the tag handling cludge.

PR #254, opened 2016-01-29
Set default binSize=10 for most of the tools

kilpert:
The default binSize for many of the tools is 50 (e.g. for bamCoverage). I suggest a smaller binSize, because it should be more in the same range with the default of computeMatrix.
The point I am trying to make is: A bigWig created with bamCoverage (default binSize=50) which is subsequently processed with computeMatrix (default binSize=10) will barely benefit from the smaller 2nd bin size. 
It's just choosing reasonable defaults. Unfortunately, a smaller bin size also increases file size!!!
I leave this open for discussion.


dpryan79:
This seems reasonable, imho.

PR #255, opened 2016-01-29
Better plot coverage image

fidelram:


fidelram:
Locally the failing test pass. I don't know what is the issue. Any ideas?

bgruening:
Could be a different matplotlib version in travis and yours?

dpryan79:
Have a look at `deeptools/test/test_plotCoverage/plotCoverage_default.svg`. The test failing is a good thing!

fidelram:
The image itself looks shifted from the viewing area. But that was true for the previous image as well (I used the linux `eog` programm, maybe it looks different with other program but that is other issue) .

dpryan79:
It's not just shifted, the left half is all over the place and outside the bounds of the image itself! I should note that matplotlib has some bugs when it produces SVG images, perhaps this is related.

dpryan79:
Here's the image that I see: http://i.imgur.com/Emz7jp6.png

fidelram:
The problem with the image is probably caused by an issue with matplotlib and is unrelated although annoying.

However, I think I found the problem: 
 * If I run a single test (`test/test_plotCoverage.py`) the test succeeds
 * If I run the tests on a batch with other tests, then the plotCoverage test fails.

Thus, I suppose that some internal matplotlib variable is changed during one of the tests that modifies the svg output.

I will try to find a workaround for this issue.

dpryan79:
I had noticed that as well. Have you tried making a png and seeing if that at least looks normal? If so, then we can chalk this up to a matplotlib bug and just ignore it.

fidelram:
@dpryan79 .png image looks correct. It is only the svg.

Meanwhile the test continues to fail. I tried using the matplotlib testing which is exactly what we need here but is all messed up in the 1.5.0 release (https://github.com/matplotlib/matplotlib/issues/5487) 

There is suppose to be a way to install `matplotlib-tests` with conda but didn't work for me. I think matplotlib 1.5.1 is needed.

I will try other workaround.

PR #260, opened 2016-01-31
added --samplesNames

fidelram:


fidelram:
@bgruening please check.

deeptools/heatmapper.py
```diff
@@ -562,7 +562,7 @@ def matrix_from_dict(self, matrixDict, regionsDict, parameters):
         self.lengthDict = OrderedDict()
         self.matrixAvgsDict = OrderedDict()
 
-    def read_matrix_file(self, matrix_file, verbose=None, default_group_name='label_1'):
+    def read_matrix_file(self, matrix_file):
```
fidelram: Those parameters were not being used in the function.

deeptools/test/test_heatmapper.py
```diff
@@ -100,14 +100,15 @@ def test_plotHeatmap_scale_regions(self):
 
     def test_plotHeatmap_multi_bigwig_pergroup(self):
         if self.run_image_tests:
-            args = "-m {}/master_multi.mat.gz --perGroup --outFileName /tmp/_test.svg".format(ROOT).split()
+            args = "-m {}/master_multi.mat.gz --perGroup --samplesLabel file1 file2 file3 file4 " \
+                   "--outFileName /tmp/_test.svg".format(ROOT).split()
```
fidelram: Added a test for `--samplesLabel` that was missing

fidelram:
I ended up changing lot of different files after realizing that the input for `--samplesLabel` was separated by spaces while for `--regionsLabel` was separated by commas. Now both are separated by space.

This PR should also solve the problem in galaxy for plotProfile and clustering that requires that the regionsLabel is the same size as the number of clusters. The default values cause an error. Thus, the default now is 'no value'.

@bgruening would not be better to add a macro for the regions and samples label? that macro can also include plotTitle that is also common between plotHeatmap and plotProfile.

vivekbhr:
Just want to add that there is is --labels option in plotCorrelation, and --sampleLabels in plotHeatmap, which both do the same thing. Maybe we should have one consistent name..

fidelram:
@vivekbhr the difference with plotHeatmap and plotProfile is that there are labels for regions and for samples, hence the distinction.

galaxy/wrapper/plotProfiler.xml
```diff
@@ -120,10 +126,21 @@
                     <option value="se">add standard error</option>
                     <option value="overlapped_lines">overlapped lines</option>
                 </param>
-                <param argument="--regionsLabel" type="text" value="coverage" size="30"
+                <param argument="--samplesLabel" type="text" size="30"
+                    label="Labels for the samples (each bigwig) plotted"
+                    help="The default is to use the file name of the sample. The sample labels should be separated by
+                    spaces and quoted if a label itself contains a space E.g. label-1 &quot;label 2&quot;" />
```
bgruening: e.g.

bgruening:
:+1:

galaxy/wrapper/plotProfiler.xml
dpryan79: ", e.g., " to be more explicit. Good catch btw!

PR #271, opened 2016-02-03
I mucked up the kmeans and hclust syntax before, so neither would work

dpryan79:
Fixing bugs I made in the wrappers. And this is why we test...


dpryan79:
I should mention that these aren't all my bugs :)

dpryan79:
This can get merged once travis is happy

galaxy/wrapper/tool_dependencies.xml
```diff
@@ -1,9 +1,71 @@
...
+        <repository name="package_python_2_7_matplotlib_1_4" owner="iuc" prior_installation_required="True" />
+    </package>
+    <package name="scipy" version="0.14">
+        <repository name="package_python_2_7_scipy_0_14" owner="iuc" prior_installation_required="True" />
     </package>
-    <package name="deepTools" version="2.0">
-        <repository name="package_python_2_7_deeptools_2_0" owner="iuc"/>
+    <package name="pysam" version="0.8.3">
+        <repository name="package_python_2_7_pysam_0_8_3" owner="iuc" prior_installation_required="True" />
+    </package>
```
bgruening: Why is this change needed? I would prefer the old version? Can I revert this change?

dpryan79: I don't think this file was even being used before. Frankly, we can probably just delete it. I just updated it to match what was actually being used by the deepTools package.

PR #274, opened 2016-02-05
Plot profile hclust

dpryan79:
Allow plotProfile to pay attention to `--hclust`. The option has been there for a while, but has been ignored. As best as I can tell this works, however I never get more than one cluster.


dpryan79:
Oh, this fixes builds on TravisCI, where it seems that the conda numpy recipe on Linux is screwed up.

PR #282, opened 2016-02-10
fix #280

### LINKED ISSUE #280: plotHeatmap with hierarchal clustering shouldn't die on fewer clusters
Currently, if the --hclust option is used an one cluster gets no members then the BED file is written but plotHeatmap then completely fails. This failure to create the plot should be fixed. plotProfile seems to be OK with this, strangely.

Additionally, we need to add a bigger warning to the galaxy wrapper about this option.

dpryan79:


dpryan79:
@vivekbhr Go ahead and remove the WIP and merge if this solves the problem.

fidelram:
In galaxy, any warning is printed, I think, after the job has completed (or failed). If this is the case a user can wait for hours without knowing what is going on.

PR #286, opened 2016-02-15
Try to fix #183

### LINKED ISSUE #183: Include Galaxy wrapper testing in travis
Do we want this? The only disadvantage I can see is that testing keeps longer.

dpryan79:


dpryan79:
N.B., this increases each run of TravisCI to ~16 minutes. I wouldn't want to push it much more than that.

dpryan79:
Note that there are issues here with database locking. Sometimes the sqlite database will be locked when a test completes and that will cause a failure (even if there was no error).

dpryan79:
Ah, that race error is probably not one, but instead related to this: http://beets.io/blog/sqlite-nightmare.html

.planemo.sh
```diff
@@ -0,0 +1,25 @@
...
+source activate deeptools_galaxy
+conda install -c bioconda samtools
+git clone --depth 1 https://github.com/galaxyproject/galaxy.git clone
+cd clone
+./scripts/common_startup.sh --skip-venv --dev-wheels
+cd ..
+pip install . 
+/home/travis/build/fidelram/deepTools/foo/bin/planemo test --galaxy_root clone --test_data galaxy/wrapper/test-data/ --skip_venv \
+galaxy/wrapper/bamCompare.xml \
+galaxy/wrapper/bamCoverage.xml \
```
bgruening: You could simply write `galaxy/wrapper` and to be save use the `-r` argument.

bgruening:
This looks awesome!

.planemo.sh
dpryan79: What does the -r option do? One issue here is that we can't test all of the wrappers, since planemo doesn't seem to support custom data types (yet).

bgruening: recursive, we can simply add this datatype with sed to the datatypes_conf ... I will look at it.

dpryan79: Cool, that'd be ideal!

PR #287, opened 2016-02-15
Fix #269, by allowing plotCoverage and plotFingerprint to accept sing…

### LINKED ISSUE #269: plotCoverage in Galaxy currently _requires_ more than 1 BAM files as input
screen shot 2016-02-02 at 12 12 04 pm
I don't think that's a meaningful requirement. Currently it simply fails when less than 2 BAM files are selected.

dpryan79:
…le BAM files, while still needing 2+ with multiBamSummary

I'll test this on the test server before removing WIP.


bgruening:
:+1: Also macros do support inheritance but this is good enough for now.

PR #290, opened 2016-02-16
Handle labels and comment lines in BED files better

### LINKED ISSUE #288: Comment lines at the beginning of BED files are incorrectly interpreted as group labels
If a BED file starts with a "#", that line is interpreted as a group label and computeMatrix will fail, since the number of labels won't match the number of groups.

dpryan79:
This fixes #288, which includes the following:

1. Comment lines at the beginning of files
2. Adjacent label lines
3. Files with label lines in the middle, but not one at the end

All of these caused errors before.

### DIFF
--- a/deeptools/heatmapper.py
+++ b/deeptools/heatmapper.py
@@ -794,8 +794,6 @@ class heatmapper(object):
         for ginterval in bed_file:
             if ginterval.line.startswith("track") or ginterval.line.startswith("browser"):
                 continue
-            if ginterval.line.startswith('#') and totalintervals == 0:
-                continue
 
             if ginterval.line.startswith('#'):
                 # check for labels with no associated entries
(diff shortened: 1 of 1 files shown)


deeptools/heatmapper.py
```diff
@@ -786,16 +786,23 @@ def get_regions_and_groups(regions_file, onlyMultiplesOf=1,
...
         for ginterval in bed_file:
-            totalintervals += 1
             if ginterval.line.startswith("track") or ginterval.line.startswith("browser"):
                 continue
+            if ginterval.line.startswith('#') and totalintervals == 0:
+                continue
 
             if ginterval.line.startswith('#'):
+                # check for labels with no associated entries
+                if groupintervals == 0:
```
fidelram: I think this case should contain the previous one on line 797, thus I think that 797 and 798 are not necesary

dpryan79: True, good catch

dpryan79:
Closing and reopening to trigger the travis test.

PR #293, opened 2016-02-16
Planemo test

dpryan79:
This adds testing of all of the Galaxy wrappers, since it adds the datatype needed by the plot* functions.

Note that some of the tests need to be updated...


dpryan79:
@bgruening: Thanks for the `sed` idea, it seems to work.

PR #295, opened 2016-02-19
Wrapper fixes

dpryan79:
It turns out that any of the wrappers with a "Background" section were rendering incorrectly. For some reason, the "Background" section kept getting rendered at the top of the page!?!

I have no clue why this was happening, it's presumably a bug in how Galaxy renders things.

### DIFF
--- a/galaxy/wrapper/bamPEFragmentSize.xml
+++ b/galaxy/wrapper/bamPEFragmentSize.xml
@@ -66,13 +66,13 @@
 <![CDATA[
 
 What it does
----------------
+------------
 
 This tool samples a BAM file with paired-end data to estimate the fragment length distribution.
 Properly paired reads are preferred for computation, i.e., unless a region does not contain any concordant pairs, discordant pairs are ignored.
 
 Output
-----------
+------
 
 The **default** output is a simple summary statistic for the observed fragment lengths.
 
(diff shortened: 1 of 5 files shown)


dpryan79:
The affected wrappers are `computeGCBias`, `plotCorrelation`, `plotFingerprint`, and `plotPCA`.

bgruening:
I fixed one of those yesterday: https://github.com/fidelram/deepTools/commit/e907472e6c8092c334c1d664a6891e01603f6a67

This is because titles are converted to div ids and these are overlapping with a Galaxy div ID that is called background. We have discussed to prefix all internal tool IDs with `_` or similar.

dpryan79:
Ah, that explains why that one was fixed when I installed it locally :)

I just copied the title changes over to the other tools

bgruening:
Magic! Sorry was to lazy to grep for other occurrences, I should have done so.

dpryan79:
No worries, I'll merge this once Travis is happy. Could you then update the other 3 wrappers in the toolshed?

bgruening:
Sure I have an other feature request for the wrappers I'm working on that I wanted to include once tested and then I will upadate everything

dpryan79:
Sounds like a good plan!

PR #297, opened 2016-02-19
Fix108

### LINKED ISSUE #108: computeMatrix on scale-regions could be unscaled nearby start/end
Hi,

some paper show metagene profile with scaled region and unscalled region in the body of region like this figures from this paper (Kruesi, et al. “Condensin Controls Recruitment of RNA Polymerase II to Achieve Nematode X-Chromosome Dosage Compensation.” eLife 2 (January 18, 2013): e00808. doi:10.7554/eLife.00808.): 
metagene_partially_scaled_body

I think it should be a good approach to see some effect because you can see "true" signal around TSS and TES and a scaled vision of the center of body all in one figure. In this review (Ferrari, et al. “Transcriptional Control of a Whole Chromosome: Emerging Models for Dosage Compensation.” Nature Structural & Molecular Biology 21, no. 2 (February 1, 2014): 118–25. doi:10.1038/nsmb.2763.) autors explain (in box3) why this kind of visualisation is better than a "simple"  scaled version.

I think if you can implement this and the "rna-seq mode" (which join all exon) then deeptools will be an essential and powerfull tool.

Many thanks to your work.

dpryan79:
Fix #108 in the next release


deeptools/plotHeatmap.py
```diff
@@ -67,6 +67,14 @@ def get_heatmap_ticks(hm, reference_point_label, startLabel, endLabel):
     w = hm.parameters['bin size']
     b = hm.parameters['upstream']
     a = hm.parameters['downstream']
+    try:
+        c = hm.parameters['unscaled 5 prime']
```
bgruening: c = hm.parameters.get('unscaled 5 prime', 0)

deeptools/plotHeatmap.py
```diff
@@ -67,6 +67,14 @@ def get_heatmap_ticks(hm, reference_point_label, startLabel, endLabel):
...
+    except:
+        c = 0
+    try:
+        d = hm.parameters['unscaled 3 prime']
```
bgruening: d = hm.parameters.get('unscaled 3 prime', 0)

deeptools/plotHeatmap.py
```diff
@@ -219,10 +238,10 @@ def plotMatrix(hm, outFileName,
         total_figwidth += 1 / 2.54
     fig = plt.figure(figsize=(total_figwidth, figheight))
 
-    hm.parameters['upstream']
-    hm.parameters['downstream']
-    hm.parameters['body']
-    hm.parameters['bin size']
+    # hm.parameters['upstream']
```
bgruening: I guess this can be removed.

galaxy/wrapper/computeMatrix.xml
```diff
@@ -118,6 +120,14 @@
                             label="Distance downstream of the region end position"
                             help="If the regions are genes, this would be the
                             distance downstream of the transcription end site."/>
+                        <param name="unscaled5prime" type="int" min="0" value="0"
```
bgruening: as a general note: if you use argument="--unscaled5prime" name will automatically calculated as argument.lstrip('-') and Galaxy will add (--unscaled5prime) to the help. This saves you a little bit of typing.

Issue #237, opened 2016-01-22
Support python3


dpryan79:
Python 2.7 is getting seriously long in the tooth. I think all of the requirements work with python3 already and if I can support both with the C API then this should be simple in pure python :)

Odds are that the 2to3 converter will make everything work.

Steps needed:
 - [x] Remove bx-python
 - [X] Add deeptoolsintervals
 - [x] Add twobitreader
 - [x] Run `2to3` and test

fidelram:
I vote for this. We can start taking the steps listed here: https://docs.python.org/3/howto/pyporting.html

dpryan79:
It looks like the only stumbling block for this is bx-python. It doesn't officially support python3 and given that that was tried and removed from the bioconda recipe, fixing it may be non-trivial.

dpryan79:
Since we only use the 2bit and interval things, I can just write replacements if needed (I'd obviously prefer not to!).

fidelram:
There is a bx-python3 https://bitbucket.org/mruffalo/bx-python3/overview

dpryan79:
It might be easier to just not use bx-python. There's a `twobitreader` module that would at least take care of the 2bit stuff. That's written in pure python, so it might be a bit slow. There's also an intervaltree module that supports python 2 and 3. I might make a fork that uses those and see how the performance compares.

dpryan79:
The `twobitreader` module needs a bit of work. I have a fork of it that removes a random `print()` that will break nose testing and confuse users. Aside from that, it generally seems to work.

dpryan79:
I have one last PR on the `twobitreader` module, but otherwise that's working in Galaxy and the CLI. The author seems to be quick at accepting PRs, this issue should be closable soon.

dpryan79:
I've started converting everything to python3. The biggest stumbling block at the moment is that any test that compares a text file whose header contains a python `dict()` (e.g., computeMatrix) will typically fail due to python3 having randomly ordered dictionaries.

dpryan79:
This is essentially working now. The only thing failing are some of the image comparisons under python3 (these differ between OSX and Linux).

dpryan79:
Shrug, sorting out the image comparison thing in python3 will be a large time sink. I'm just disabling testing of that component there, since the failing tests appear to produce identical images and the diffs mostly look like reordered lines or differences in floating point precision (e.g., an extra digit at the end of a coordinate).

[...]

PR #300, opened 2016-02-23
NFS empty file fix

bgruening:
This hopefully will close all open file handles which is essential on slow file systems like NFS. Otherwise the file will stay empty in Galaxy.


bgruening:
Lessens learnt don't fix this with your cellphone without pep8 linting ;)
I got PR #300!!!!

fidelram:
I will try to close always from now on

dpryan79:
Hopefully this solves the issue.

dpryan79:
BTW, I'm amazed that we've never run into this, since all of our mounts are via NFS.

bgruening:
Thanks for merging!

dpryan79:
This is an important enough fix that I'll bump #237 to version 2.3, since the other things for the 2.2 milestone are essentially already done.

PR #308, opened 2016-03-02
Wrapper update 2.2

bgruening:


galaxy/wrapper/test-data/profiler_result2.tabular
```diff
@@ -0,0 +1,3 @@
+bin labels		-0.0Kb	0.0Kb
```
bgruening: @dpryan79 any idea why profiler and heatmapper produces the same additional data?

dpryan79: They probably shouldn't, that would seem to be a bug. I'll look into that tomorrow.

dpryan79: I have a hotfix branch that fixes the problem you found previously. We can just add fixes to that, since it'll be 2.2.1.

bgruening: so sorry, but I haven’t had time until know :(

dpryan79: No worries, I was talking about fixes for the bugs you've found while adding more galaxy tests.

dpryan79: It turns out that the `--outFileNameData` option is returning the data for the profile on the top, while `--outFileNameMatrix` is returning the data for the heatmap on the bottom.

dpryan79: I guess we have three options:
 1. Remove the `--outFileNameData` option from plotHeatmap, since `--outFileNameMatrix` is really already doing what it should.
 2. Modify how the function is called in plotHeatmap so it saves the heatmap data, rather than the profile data.
 3. Change the documentation to note this.

bgruening: 1. or 2. the only usage I have seen so far is in using the matrix not the other file.

dpryan79: I'll go with option 1 then. I don't really see the point of the argument, it only exists because the arguments are partly shared with `plotProfile` :P

dpryan79: cf. https://github.com/fidelram/deepTools/commit/adfe2712222d62e18cb94f4fc6053045f008dcbe

Let me know if you find anything else.

bgruening:
I'm ok with merging if the tests are passing. Thanks!

bgruening:
Tests are passing!

dpryan79:
Merged, nice work!

bgruening:
Please ping me with the new release. I will create some new Galaxy packages and push this to the TSs. Thanks!

dpryan79:
Will do, presumably we'll push out 2.2.1 on Monday (unless someone finds another important issue in the interim).

PR #313, opened 2016-03-07
fixed problem with hierarchical clustering that reported less clusters

fidelram:
This was due to a different convention on naming clusters compared to kmeans


fidelram:
I am creating a new pull request based on branch develop, this one was based on branch master

PR #322, opened 2016-03-22
Add heatmap as option for plotProfile

bgruening:


dpryan79:
I guess we forgot that option :P

bgruening:
I guess so too ;)

PR #323, opened 2016-03-23
improved visualizations and documentation for plotHeatmap and plotProfile

fidelram:
now the visualization are based on log2ratios which should be best practice when dealing with ChIP-seq data.


fidelram:
Any idea why the test fails?

dpryan79:
It's an issue with the version of sqlite that comes with conda. Ignore it. If you run Travis again it'll probably pass. This happens about 25% of the time and is likely due to sqlite not being compiled to nanosleep.

Issue #76, opened 2015-04-27
add possibility to use bed12 or gtf to construct metagene heatmap and profile


lelouar:
add the possibility to join all the exons for construct transcript genebody with RNA-seq data

fidelram:
Commit 738a840069e8435625874b48dee3b1244b2c50e1 initially address the issue by using a custom bed file reader. This is just one step into the solution of the issue.

dpryan79:
It turns out that a side-effect of implementing this (I'm working on it and #71 now) will be to largely get rid of bx-python. That'll then make #237 easier :)

dpryan79:
This is now implemented in the `develop` branch, which will become the 2.3 release.

Issue #71, opened 2015-01-28
bamCorrelate: gtf as --BED-file input


adomingues:
Is possible to use a GTF file as an argument for the bamCorrelate?

The reason that I am asking, is that frequently gene annotations come in that format. For instance, I am comparing bams form 2 sequencing runs mapped with the same GTF. I would make sense to use that GTF for the comparison of the sequencing runs.

Of course I can easily convert the GTF in bed, but what is the fun in that? :)

dpryan79:
This is now implemented in the `develop` branch, which will become the 2.3 release.

PR #325, opened 2016-03-29
Feature/deeptoolsintervals

dpryan79:
This implements #71 and #76, which results in almost entirely removing bx-python. Given the amount of change involved, it'd be good if we use this for a while internally and then add a "deepTools 2.3 beta" to our public Galaxy instance before making a real release.


.travis.yml
```diff
@@ -28,9 +28,10 @@ install:
 
 # command to run tests
 script:
-  - flake8 . --exclude=.venv,.build,foo --ignore=E501,F403,E402
-  - nosetests --with-doctest --ignore-files=plotCoverage deeptools
-  - nosetests deeptools/test/test_plotCoverage.py
+  - flake8 . --exclude=.venv,.build,foo,build,deeptoolsintervals/__init__.py --ignore=E501,F403,E402
```
bgruening: foo, build?

deeptools/bamCompare.py
```diff
@@ -186,13 +190,18 @@ def get_scale_factors(args):
                     (scalefactors_dict['sites_sampled'],
                      args.sampleLength)
 
-                print "size factor if the number of mapped " \
+                print "ignoring filtering, size factors if the number of mapped " \
```
bgruening: At some point we should get rid of these print statements.

deeptools/bamCompare.py
```diff
@@ -186,13 +190,18 @@ def get_scale_factors(args):
...
                 print tuple(
                     float(min(bam1.mapped, bam2.mapped)) / np.array([bam1.mapped, bam2.mapped]))
 
         elif args.scaleFactorsMethod == 'readCount':
-            scale_factors = float(min(bam1_mapped, bam2_mapped)) / np.array([bam1_mapped, bam2_mapped])
+            args.bam = args.bamfile1
+            args.scaleFactor = 1.0
+            s1 = get_scale_factor(args)
+            args.bam = args.bamfile2
+            s2 = get_scale_factor(args)
```
bgruening: I guess a cleaner solution would be to give the bam file as second parameter to get_scale_factor()

.travis.yml
dpryan79: `build/` is created with `python ./setup.py install`. `foo/` is a venv for the planemo tests.

deeptools/bigwigCompare.py
```diff
@@ -51,17 +51,19 @@ def parse_arguments(args=None):
                         required=False)
 
     parser.add_argument('--ratio',
```
bgruening: This is a shared one, with e.g. bamCompare.

deeptools/countReadsPerBin.py
```diff
@@ -269,6 +270,18 @@ def run(self):
             # in case a region is used, append the tilesize
             self.region += ":{}".format(self.binLength)
 
+        # Handle GTF options
+        transcriptID = "transcript"
+        exonID = "exon"
+        transcript_id_designator = "transcript_id"
```
bgruening: Can we put this block into a private function as utils, this at least two times.

dpryan79: I suppose so.

dpryan79:
Closing due to accidentally targeting to `main`. Anyway, I'll just `git flow feature finish` this.

PR #331, opened 2016-04-05
Fixing quotes of multiple filenames in Galaxy

bgruening:
Ones and for all, I added a text file with a space in the name ;)


galaxy/wrapper/multiBamSummary.xml
```diff
@@ -16,8 +16,8 @@
             @THREADS@
 
             --outFileName '$outFile'
-            --bamfiles '#echo "' '".join($files)#'
-            --labels '#echo "' '".join($labels)#'
+            --bamfiles #echo " ".join($files)#
+            --labels #echo " ".join($labels)#
```
dpryan79: I assume we no longer need the `#` at the end of each of these lines.

bgruening: It more or less closes the cheetah template. It could be removed yes.

dpryan79:
lgtm!

bgruening:
Thanks Devon!

PR #340, opened 2016-04-15
problem with heatmapper scale regions

fidelram:


fidelram:
@devon the develop version of deepTools computeMatrix produces a heatmap that is different from master. This is very noticeable once is plotted. The tests were previously been passing because the wrong 'master_scale_reg.mat' file was being used.

dpryan79:
`master_scale_reg.mat` (both uncompressed and gzipped) is identical between `develop` and `master`.

dpryan79:
In fact, the file has been largely unchanged since 2013. Apparently the shell script was never actually used to generate the file.

fidelram:
@devon actually what I found is that the unzipped version of master_scale_reg.mat didn't match the gzipped version. Using the wrong version, the test did pass. 

In this branch I added the corresponding unzipped version and updated the script that generates the gzipped file which in the test uses a bin size of 1 and the script was using a larger size.

Since the tests used the original master_scale_reg.mat.gz to produce the images all tests passed. But if master_scale_reg.mat.gz is generated from the development version the output is different than from master. Visually it looks as follows, the left if the master version, the right is the develop version:

out

dpryan79:
Ah, so it's the generation of the matrix file that's messed up. OK, I'll have a look.

dpryan79:
I have a fix of this getting tested by Travis now. The problem was caused whenever a region was smaller than the number of bins requested. In those cases the start and end indices into the values array were the same, so `[]` was being appended to the value array. That was a silly mistake on my part.

When Travis is done I'll rerun the PR testing, since it'll presumably pass then.

dpryan79:
Travis is now happy (a galaxy test is failing due to the annoying sql locking issue, but that can be ignored). Presuming that this PR is finished then go ahead and merge.

PR #345, opened 2016-04-19
updated galaxy wrappers for multi color heatmaps and dpi

fidelram:


dpryan79:
Guess I used the wrong email address in the config in the docker container.

Anyway, the options are now all working in Galaxy. If you're otherwise happy with things then remove WIP and we can merge once Travis gives the :+1:

PR #362, opened 2016-05-23
Fix ignore for normalization

fidelram:
this PR fixes three issues:

In branch develop (but not in master) the computation of scaled factors for bamCompare was not working. We didn't have tests to check if the scale factors were being computed correctly. Now, I added new test to check for this under different conditions including ignoreForNormalization and filtering.

While working on the tests I found that when writing a bedgraph file sometimes the last value to be stored from a list assigned to a thread was skipped if this was a zero. Also, the precision of this last value was 1 digit instead of 2 as all other values.


fidelram:
Test are failing because wheels.galaxy.org is not responding

bgruening:
Working on it!

fidelram:
I am currently updating the galaxy tests

PR #363, opened 2016-05-25
add feature that allows to select a subset of samples used for sorting

steffenheyne:
within each group, add option --sortUsingSamples, provided samples are numbered from 1, numbers are checked if they are in range of existing samples. Nothing is changed yet in the matrix header line.

### DIFF
--- a/deeptools/computeMatrix.py
+++ b/deeptools/computeMatrix.py
@@ -247,6 +247,13 @@ def computeMatrixOptArgs(case=['scale-regions', 'reference-point'][0]):
                           choices=["mean", "median", "max", "min", "sum",
                                    "region_length"],
                           default='mean')
+    
+    optional.add_argument('--sortUsingSamples',
+                          help='List of sample numbers (order as in matrix), ' 
+                          'that are used for sorting by --sortUsing, '
+                          'no value uses all samples, '
+                          'example: --sortUsingSamples 1 3',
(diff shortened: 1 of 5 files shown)


steffenheyne:
maybe someone can review ist? thanks!

deeptools/computeMatrix.py
```diff
@@ -370,7 +377,18 @@ def main(args=None):
     scores_file_list = args.scoreFileName
     hm.computeMatrix(scores_file_list, args.regionsFileName, parameters, blackListFileName=args.blackListFileName, verbose=args.verbose, allArgs=args)
     if args.sortRegions != 'no':
-        hm.matrix.sort_groups(sort_using=args.sortUsing, sort_method=args.sortRegions)
```
fidelram: I think here is args.sortUsingSamples instead of args.sortUsing

dpryan79:
At some point it's easiest to do stuff like this after the fact. Have a look at the feature/riboseq_352 branch. There, I have a script (`scripts/subsetMatrix`) that does things like subset/merging matrices from computeMatrix. I foresee us making a new tool that only manipulates the output of computeMatrix.

fidelram:
Seems to be OK. @steffenheyne I assume that you have run the code and that it gives you correct results.

steffenheyne:
yes, output make sense to me.

PR #375, opened 2016-06-15
print cluster labels without space

vivekbhr:
The space for heatmapper cluster labels (cluster 1) was annoying. so I changed it to underscore (cluster_1)


fidelram:
The fact that the tests passed means that we don't have yet a test that involves clusters.

PR #381, opened 2016-06-24
Filterout option has gone obviously

bgruening:


bgruening:
Oh this failure are actually not from this PR as it seems. Merging?

dpryan79:
Yeah, sorry for not catching that! PEP8 got updated a month ago and everything broke. I have this fixed in `develop` already. If you have a chance, can you make the same PR against that branch? Then the change will be there for the next version.

PR #391, opened 2016-07-20
I got the bamCoverage wrapper code in the bamCompare wrapper instead :(

dpryan79:


dpryan79:
@bgruening: Note that I made a mistake in the Galaxy wrapper that this should fix.

PR #397, opened 2016-08-05
added new option ''--refPointLabel' for computeMatrix reference-point…

kilpert:
…; refPointLabel, startLabel, and endLabel are now written to the header line of computeMatrix if applicable


dpryan79:
FYI, the Travis tests are going to fail because because the header is part of the testing. Just update the test files to include this and everything should go green :)

PR #462, opened 2016-12-22
fixed bug

sklasfeld:
line 165 caused an error when you used the "-l" parameter to label the samples. 
TypeError: list indices must be integers, not list

Therefore, I changed the line so that it now works!


dpryan79:
Good catch, thanks for the fix! This will now be corrected once 2.4.2 comes out.

Issue #464, opened 2017-01-04
Galaxy specific - sortRegions and sortUsing 


MoHeydarian:
I'm using deepTools 2.4 in Galaxy. I would like to plot my data with computeMatrix -> plotHeatmap in  the order that the regions are provided to deepTools. There are options to not sort the data (--sortRegions 'no'), though when the advanced options are revealed in the tool form the --sortUsing parameter is invoked and the initial order of the regions is lost. When the advanced options are not revealed the initial order of regions is preserved.

Can the sortUsing parameter be excluded/ignored when invoking --sortRegions 'no', or is there a better way (or am I missing something)? 

Thanks in advance!

dpryan79:
At least in the 2.4.1 wrapper the default is 'keep', which is labeled "maintain the same ordering as the input files". If you specify "no" ("no ordering") or "keep" ("maintain the same ordering as the input files") then --sortUsing is ignored regardless of what is specified.

MoHeydarian:
Yes I see that specifying "no ordering" or "maintain the same ordering as the input files" for computeMatrix and using plotHeatmap without revealing the advanced options preserves the order. 

When revealing the advanced options of plotHeatmap the order is lost, even when specifying "no ordering" for the sortRegions option. I can share the data I'm working with, if that helps.

bgruening:
Mo, maybe this can fix it: https://github.com/fidelram/deepTools/issues/464

MoHeydarian:
Would that revision hide the sortUsing selection box in the tool form if sortRegions is set to "no ordering" (in plotHeatmap)? If the user selects "no order" for sortRegions it could be confusing to still show the selection menu for sortUsing.

dpryan79:
Yes, can you share an appropriate history or dataset with me (I'm ryan@ie-freiburg.mpg.de on the main usegalaxy.org site). If correct, this is a bug in more than just the Galaxy wrapper, since --sortUsing should always be ignored if --sortRegions is either not given or "no" or "keep".

MoHeydarian:
Usegalaxy.org is still on version 2.3.6 (which also doesn't maintain the order of the regions), I'll share a history with you on my Cloudman instance by email shortly.

dpryan79:
Thanks for the cloudman instance, I can see the issue and will try to get it fixed when I'm back in the office on Monday (I'll push out version 2.4.2 then).

dpryan79:
[...]

PR #465, opened 2017-01-05
464 fix

bgruening:


bgruening:
It's late but this can fix https://github.com/fidelram/deepTools/issues/464

dpryan79:
#464 would appear to be a plotHeatmap bug, wherein `--sortRegions no` is being ignored in it (I'm not sure how this could happen). With the exception of the deeptools_macros change, the other code is already present in the plotHeatmap wrapper.

PR #469, opened 2017-01-19
Fix wrappers multiple inputs

dpryan79:
Ping @bgruening: This fixes a bug in the wrapper for multiBamSummary and multiBigwigSummary if people specify that the want to manually set the order of files given to the command. I guess we didn't sufficiently test this (thanks to `planemo serve` I've now done so).


bgruening:
to late but here my :+1: :)
You are right, we are not testing this condition :(

PR #497, opened 2017-03-24
Use `element_identifier` to label samples according to their name in …

mvdbeek:
…the history

With this change you get the history name or the collection elements' element_identifier as label (unless overriden with a custom name). This is really helpful for us when we have a bunch of samples in a collection, we don't have to be careful to not mislabel samples.


fidelram:
@bgruening Can you take a look at this PR

bgruening:
Already gave my plus one. Looks good. And can be merged. Thanks Marius.

PR #499, opened 2017-03-29
Element identifier

mvdbeek:
Here are some more instances where element_identifier will result in better labels.


dpryan79:
I'll take a look at plotPCA today. Thanks for the PR, btw, the changes are great!

dpryan79:
For some reason, `multiBamSummary` is returning a matrix that's all the same number, so the PCA breaks.

dpryan79:
I've forked this PR and merged it in after making some changes to how the tests work and what the expected output should be. Thanks!

PR #500, opened 2017-03-30
Mvdbeek element identifier

dpryan79:


mvdbeek:
Great, thanks a lot @dpryan79 ! Can you notify @martenson or @davebx once this goes to the toolshed?

dpryan79:
Sure, version 2.5.0 is going out tomorrow, so it'll be after that's available in bioconda and I have a chance to make a PR against the IUC.

PR #511, opened 2017-04-19
fixed debug global bug and shutil bedgraph join bug in correctGCBias

drakeeee:
I received two errors when running correctGCBias using python3.5 on ubuntu14.

This PR addresses those 2 errors:
1) The debug global is not assigned if the script is run without using the Tester Class. I added a try/except statement to catch it.

2) The shutil.copyfileobj method needs binary files for both the source and destination file. I modified this accordingly.

All appears to be in working order after these modifications. Thanks for the great work.

--Adam
adam.drake@freenome.com


dpryan79:
Thanks Adam! This will be in the 2.5.1 release, which will hopefully go out rather soon.

PR #557, opened 2017-07-12
Use matplotlib `compare_images` for testing instead of svg method

fidelram:
Compare_images should be more stable across platforms and matplotlib versions. 
I updated the requirements to matplotlib 2


fidelram:
@bgruening @dpryan79 Galaxy tests are failing. I think this is because of the matplotlib version now used (2.0.2) in contrast to 1.5 before, but I don't know how to update the galaxy test images. All other tests are passing.

One significant difference with matplotlib 2 is the use of nicer default colors for lines and dots.

Shall we update the galaxy images or roll back to matplotlib 1.5 ?

dpryan79:
The Galaxy testing already uses matplotlib 2.X since it's using conda.

dpryan79:
In fact, Galaxy isn't comparing the images, just the file sizes. So we either need to update the files it compares against (`galaxy/wrapper/test-data/`) so the sizes are within the delta tolerance or just increase the tolerance on the failing tests. My preference would be the former. I can run planemo locally real quick on that branch and then update the images with what Galaxy is actually producing.

dpryan79:
@fidelram Was there anything else you were planning to do with this branch or is it ready for merging?

PR #575, opened 2017-08-08
plotEnrichment: also print counts in outRawCounts

steffenheyne:
see https://github.com/fidelram/deepTools/issues/571


dpryan79:
I've fixed the PEP8 issues, though I assume the Galaxy tests will need to be updated too.

dpryan79:
I've updated the Galaxy test. Assuming that passes I'll go ahead and merge.

dpryan79:
This is now merged into `develop`.

PR #586, opened 2017-08-18
properly quote filenames as well

bgruening:


dpryan79:
Can you merge this into `develop` as well?

bgruening:
For me it is ok if it hits the next release. Any reason why this needs to be in develop?
Or do you want to have it only in develop?

dpryan79:
If you put it in master then it will never see the next actual release. Release only come from the `develop` branch. The master branch holds only released stable code (I follow the `git flow` methodology on my projects, since people have a habit of installing directly from github and I don't want them to get non-functional code unless they know what they're doing).

bgruening:
It's now targeting the develop branch, but I'm not sure I understand. `develop` is merged into master before a release and you release the `master` branch right? At this time the change would appear in a new release.

Have a nice weekend!
Bjoern

dpryan79:
Release branches come off of `develop` and are merged into master. The releases themselves come from release branch and not master itself.

PR #588, opened 2017-08-18
--normalizeUsing CPM, BPM, RPKM

vivekbhr:
Since RPKM gives ugly values at small bin sizes, I added `--normalizeUsingCPM` . Wanted to make `--normalizeUsingBPM` which would have been same as TPM (per bin), but was difficult as it needs sum of coverages of all bins for calculation of norm factors, but in current implementation, scale_factors [are multiplied to](deepTools/deeptools/getScaleFactor.py) each bin, and extracting sum of coverage across all bins doesn't seem trivial..


vivekbhr:
yey! All green..

I have changed the option --normalizeUsingRPKM to `--normalizeUsing <option>`, which takes the values 'RPKM',  'CPM' or 'BPM' . CPM and BPM (in principle) also work for bamCompare. I have not tested the accuracy of the result though..

dpryan79:
Great! Sounds like the last things to do are:

 - [ ] Confirm this works for bamCompare and produces useful results (documenting if not and possibly removing the option from it in that case)
 - [ ] Update readthedocs
 - [ ] Modify the Galaxy wrappers

vivekbhr:
:see_no_evil:

dpryan79:
I'll add this to the 2.6 release milestone, since I imagine you'd like it to be available for that release.

dpryan79:
@vivekbhr If you test `bamCompare` I'll deal with Galaxy and readthedocs.

vivekbhr:
I think it would be faster to do the other way.. I was testing bigwigCompare and since the zero bins get merged the two bigwigs produce different coordinates so it's difficult to test whether for a particular bin, the operation worked.

vivekbhr:
Ok this is what I did..
- Removed the whole subtract special code from bamCompare and clarified the two modes (scaling and normalization). Scaling (readcount/SES) can be turned off and per sample norm can be applied before taking the ratio instead.

- Added an alternative name for --ratio (calling it --method), since --ratio is counter intutive --ratio is still the default.

- Modified galaxy wrapper accordingly.

- Modified docs inside the code. I didn't see any special place in RTD that mention this so I didn't modify RTD files.

I would appreciate if someone can double check whether bamCompare gives the same output as expected manual computation, when using --normalizeUsing option with bamCompare. It looks OK to me..

If not using --normalizeUsing, bamCompare should work like before, except that --normalizeTo1x won't work with any method (previously it worked only with subtract). Instead people should use `--normalizeUsing <option>`

vivekbhr:
travis failing the test of bamCompare and bamCoverage due to the use of `--ratio subtract`

fidelram:
Thanks @vivekbhr for this changes. Certainly the code needed some clean up and the new methods seem useful.

Given the changes in the code related to previous functionality we need to be sure that nothing was broken.  The travis tests do not contemplate all cases so we need to check that at least one test involving each method is present and compared with previous results. I noticed that @vivekbhr modified a test to add  --normalizeUsing CPM but a test for BPM may be missing.

I noticed that the CPM and BPM are only defined (using words) on the galaxy wrappers but not on the command line help of the tool. And the other way around, the math definition is missing on the galaxy wrappers.

vivekbhr:
@fidelram for a fixed bin size CPM and BPM end up giving the same result (maybe I should add this to the doc as well). So I only added the test for CPM. I added more info in command line doc but I think in galaxy the options appear as a dropdown list, so I don't know whether I should write a long description there. RPKM also didn't have log description earlier in galaxy

dpryan79:
For Galaxy, please add a small blurb at the bottom where the restructuredText is. If there isn't already a paragraph describing 1x/RPKM then add them too.

vivekbhr:
@dpryan79 planemo test is failing due to XML linting issue in bamCompare and bamCoverage .xml files.. Can you help?

dpryan79:
Given that this will break current workflows, I wonder if we should make this version 3.0.0 to make that clearer. Thoughts?

dpryan79:
@vivekbhr This is now passing on Travis. Are the Galaxy wrappers updated on this or does that still need to be done? I assume they've been updated since the tests are passing. If everything is finished then go ahead and merge.

vivekbhr:
I have updated all wrappers but I have not tested whether the modified galaxy wrappers work. Could you please test that?

dpryan79:
I'd be surprised if the planemo tests didn't break if you messed them up, but sure :)

deeptools/bamCompare.py
```diff
@@ -83,8 +83,13 @@ def getOptionalArgs():
...
                           help='Method to use to scale the samples. '
-                          'Default "readCount".',
-                          choices=['readCount', 'SES'],
+                          'If scaling is selected, per-sample depth normalization '
+                          'would not be used, instead counts would be scaled between samples '
+                          'to account for sequencing depth differences between the two. '
+                          'As an alternative, scaling can be set to None, and the option '
+                          '--normalizeUsing <method> can be used, to normalize each sample '
+                          'by sequencing depth before comparing them. ',
+                          choices=['readCount', 'SES', 'None'],
```
dpryan79: It's a bit unclear what the first sentence here actually means. Do you mean "If a method is selected, then it will be used to compensate for sequencing depth differences between the samples. As an alternative, this can be set to None and the option..."?

dpryan79:
- [x] bamCoverage
    - [x] `--normalizeTo1x`
    - [x] `--normalizeUsing RPKM`
    - [x] `--normalizeUsing CPM`
    - [x] `--normalizeUsing BPM`
    - [x] `--normalizeUsing None`
 - [ ] bamCompare
    - [ ] `--normalizeUsing RPKM`
    - [ ] `--normalizeUsing CPM`
    - [ ] `--normalizeUsing BPM`
    - [ ] `--normalizeUsing None`

`--normalizeUsing None` should still print something to `stderr`, for those who like that for logging. I'll change that and some of the wording.

deeptools/bamCompare.py
vivekbhr: yes I meant : ``` if --scaleFactorMethod is selected...... alternatively --scaleFactorMethod can be set to None........```

dpryan79: Thanks, thought I better double check before I change it :)

dpryan79:
Travis is happy, the documentation has been updated, and I've tested/fixed the wrappers locally with `planemo`, so I'll go ahead and merge. Thanks @vivekbhr !

bgruening:
Cool, great work @vivekbhr!

fidelram:
I was wondering if for the next release we should change the parameter `--ratio` to `--operation`. This could be the right time given that we are already changing the parameters. The reason for this change is that `--ratio subtract` does not make much sence. Any opinion on this @vivekbhr @dpryan79 @bgruening ?

vivekbhr:
Yes indeed.. I was thinking about it when I was doing the changes..
Also, we should be changing the name of --normalizeTo1x to `--effectiveGenomeSize` and move the `--normalizeTo1x` flag as an option `RPGC` under `--normalizeUsing` in order to be consistant with other methods..

dpryan79:
I support removing `--ratio` altogether. This PR already added `--method` as a synonym, so whether we stick with that or switch to `--operation` is nether here not there with me. I agree that we should ditch `--normalizeTo1x` and merge it into `--normalizeUsing`. Honestly, I'd make `SES` and `readCount` fall under that too, after the "wenn schon denn schon" philosophy. I'll create an issue where we can discuss this further.

dpryan79:
Further discussion can take place here: https://github.com/deeptools/deepTools/issues/629

PR #615, opened 2017-10-06
adding the new plotTracks option to deepTools.

fidelram:


vivekbhr:
As per recent internal discussion, this should be made a new package **trackPlotter**, which would then be imported by both deepTools and hiCEx

fidelram:
This is now part of pyGenomeTracks

PR #617, opened 2017-10-18
--filterRNAstrand help text

daler:
Currently the help text for `--filterRNAstrand` is:

```
--filterRNAstrand {forward,reverse}
                        Selects RNA-seq reads (single-end or paired-end) in
                        the given strand. (default: None)
```

From this it's unclear whether reads are _selected_ or reads are _selected for removal_. This minor change clarifies the behavior of the argument.


dpryan79:
FYI, a version of this is included in the 2.5.4 release that I'm pushing out now. Thanks for bringing this up!

PR #625, opened 2017-11-09
added interpolation parameter to plotHeatmap

fidelram:
This PR recovers the previous behaviour of plotHeatmap wich was to add an interpolation other than 'nearest' when the number of rows was large. A parameter to provide the interpolation method was added.


fidelram:
@dpryan79 any idea what could be wrong with image tests in linux? The matplotlib is the same version than the one used locally, yet the resulting images have slightly different size.

PR #657, opened 2018-01-26
fix a nasty python3 error

bgruening:
The following command:
```
plotCorrelation --corData test-data/multiBamSummary_result1.npz --plotFile foo.png --corMethod "spearman" --whatToPlot "heatmap"  --colorMap 'RdYlBu'  --plotTitle ''  --plotWidth 11.0 --plotHeight 9.5  --plotFileFormat "png"  --outFileCorMatrix foo2.ma
```
produces the following error in a python3 environment:
```
[b'bowtie2 test1.bam', b'bowtie2 test1.bam']
Traceback (most recent call last):
  File "/home/bag/miniconda3/envs/__deeptools@2.5.5/bin/plotCorrelation", line 11, in <module>
    main(args)
  File "/home/bag/miniconda3/envs/__deeptools@2.5.5/lib/python3.6/site-packages/deeptools/plotCorrelation.py", line 224, in main
    corr.save_corr_matrix(args.outFileCorMatrix)
  File "/home/bag/miniconda3/envs/__deeptools@2.5.5/lib/python3.6/site-packages/deeptools/correlation.py", line 170, in save_corr_matrix
    file_handle.write("\t'" + "'\t'".join(self.labels) + "'\n")
TypeError: sequence item 0: expected str instance, numpy.bytes_ found
```
I think this is due to an older version of a `/multiBamSummary` matrix that once loaded with py3 produces this error. I think its worth to have this fix in to also load old datasets. we could break this in deepTools3 is really needed.

### DIFF
--- a/deeptools/correlation.py
+++ b/deeptools/correlation.py
@@ -83,7 +83,7 @@ class Correlation:
 
             self.matrix = np.ma.compress_rows(np.ma.masked_invalid(self.matrix))
 
-        self.labels = _ma['labels']
+        self.labels = self.labels = list(map(str, _ma['labels']))
 
         assert len(self.labels) == self.matrix.shape[1], "ERROR, length of labels is not equal " \
                                                          "to length of matrix samples"


dpryan79:
Hmm, I wonder if numpy saves things differently under python2 vs python3

dpryan79:
The Galaxy tests can be ignored, redirect to develop if you really want that tested. Otherwise this looks reasonable, though I'd really like to know why the heck it's happening in the first place.

dpryan79:
I'm testing a fix for the Galaxy tests (basically, just copying over what's in the develop branch). Do you have a preference of testing Galaxy under python 2 vs 3? I always did python 2 since that's all that worked, but it looks like python 3 is the default now in other places.

bgruening:
The conda environment is by default python3 (miniconda3). So installing deeptools from Galaxy will pull down the python3 version of conda-deeptools. So I would prefer python3.

dpryan79:
I have a branch using Galaxy in python3. I notice a number of other random numpy byte arrays that need to be converted to strings. I think I'll work on this on Monday and push out 2.5.6 then.

PR #659, opened 2018-01-28
Galaxy python3

dpryan79:
This implements #657 from @bgruening and also does Galaxy tests under python 3 rather than python 2. It's rather annoying the numpy in python 3 has issues loading matrices created in python 2.


dpryan79:
The version still needs to be bumped.

PR #674, opened 2018-03-01
str() casting needs to happen to check that the value is not 0

bgruening:


dpryan79:
I'll need to implement this in `develop` and the `release/3.0.1` branch as well, since 3.0.1 is probably going to come out today.

dpryan79:
No real CI errors, just the typical database locking issues (I really wish I could get planemo to properly use postgresql on Travis...it never seems to work). I'll merge and copy this to the branches I mentioned above.

PR #699, opened 2018-05-02
Fix computeMatrix option help typo

vreuter:


dpryan79:
That's an embarrassing one, thanks @vreuter !

PR #735, opened 2018-07-05
Feature/fix734

JoseEspinosa:
The problem with the solution proposed at e4028c1c77bc3496fef57d2360a714e2d460b210 is that then the number of samples on the main plot corresponds is set to be the number of eigenvalues and thus some sample are missing from the plot. I tried to solve this by creating two separate variables `n_bar` and `n` that are set to the same value and only `n_bar` is modified when the condition `if eigenvalues.size < n:` is met.

### DIFF
--- a/deeptools/correlation.py
+++ b/deeptools/correlation.py
@@ -646,7 +646,9 @@ class Correlation:
             Wt = np.dot(m2, Vh.T).T
 
         if plot_filename is not None:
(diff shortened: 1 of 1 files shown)


dpryan79:
Good point!

dpryan79:
Thanks for reporting and fixing this!

PR #747, opened 2018-08-09
add tool/script to convert chromosome names of bigwig files

steffenheyne:

This script make use of @dpryan79 's repo for chromosome mappings. One can give a bigwig and and FROM and TO format as well as a genome, then the tool maps the possible chromosomes and writes out the new bigwig file

I don't know if we want to include this in deeptools at all, or snakePipes? I just wanted to share this somwhow. It uses pybigwig, so deeptools has this already as dep.

In SnakePipes I think of a simple option to produce also UCSC-conform tracks that would use this script

What do you think?

### DIFF
--- a/scripts/convertChromsBigWig.py
+++ b/scripts/convertChromsBigWig.py
@@ -1,0 +1,189 @@
+#!/usr/bin/env python
+#-*- coding: utf-8 -*-
+
+import sys
+import pyBigWig
+import requests
+import re
+import argparse
+from argparse import RawTextHelpFormatter
+import pprint
+import itertools
+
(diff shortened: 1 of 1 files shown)


dpryan79:
This can certainly go under scripts, though I worry that it won't get as noticed as it deserves to be. I would suggest that I additionally create a link to it from the github pages for pyBigWig and the chromosome name mappings. Would that work for you?

steffenheyne:
ah ok, then I need anyway another solution if I want to use that directly in snakePipes...

...another thing, and definitely not rocket science, but just for convenience: the script could be extended to also convert bed, gtf, bam (header)

PR #818, opened 2019-03-06
Cache resulting counts rather than just decreasing the bin size

dpryan79:
your pull request :**


dpryan79:
This should decrease run times in things like `bamCoverage` when the signal is chunky and we have to sample a large portion of the genome.

PR #876, opened 2019-09-13
[WIP] added a silhouette calculation

lldelisle:
your pull request :**

 - [X] Does the PR contain new feature?

Hi,
This is a beginning of a PR that I would like to discuss.
What I did:

- add the `--silhouette` arguments in `heatmapperOptionalArgs`.
- when this option is put, when `hmcluster` is launched, the silhouette is evaluated with `silhouette_samples` from `sklearn`. It is stored in a new attribute of the `_matrix` class.
- In the std err the mean of the silhouette is output.
- in the `save_BED` method, it will always print an additional column with the silhouette values (initialized to nan).

What is missing:

- tests

What I would like to discuss:

- This will change the bed output in all cases (even if `--silhouette` is not put). I think it would be better to change it.
- For the moment silhouette is a new attribute of `_matrix` and I agree it would make more sense in `regions` but then I am wondering if `regions` should not be a class to be sure of what is what in this array. For me it looks like, the region is defined from 
https://github.com/deeptools/deepTools/blob/df8536adf850938d557ff17811bbda88f6895ff9/deeptools/mapReduce.py#L123-L126
But this is only when a bed is provided, I do not know which are the other cases.
- Should we plot the values? A summary per cluster of the values?

Thanks


lldelisle:
Hi,
I think the code is not working... but I can propose a solution, do you want me to commit on this branch?

dpryan79:
I'm fixing it locally and will update the PR.

lldelisle:
ok perfect

dpryan79:
@lldelisle I think this should be producing more reasonable results again. Can you double check that with your own data?

dpryan79:
@lldelisle Have you had a chance to test the current PR by chance?

PR #901, opened 2020-01-03
Using --xRange and --yRange fails in galaxy due to the single quote. …

cgirardot:
Single quotes removed. 

your pull request :**

no
yes
no
yes

### DIFF
--- a/galaxy/wrapper/plotCorrelation.xml
+++ b/galaxy/wrapper/plotCorrelation.xml
@@ -19,10 +19,10 @@
             #else:
                 --plotTitle '$plotting_type.plotTitle'
                 #if str($plotting_type.xRange) != '':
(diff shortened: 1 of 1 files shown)


dpryan79:
@bgruening Is there an appropriate way to guard against code injection via parameters that are meant to take a list? Should we change this to take a min an max value for each range instead? That would allow quot

dpryan79:
Actually, maybe I can change the logic to split by ` ` and then apply `int()` in case the input isn't a list. That would allow us to guard against code injection while still allowing the wrapper to work properly.

dpryan79:
I need to test this to ensure it works properly. I'll do that tomorrow.

cgirardot:
👍

PR #931, opened 2020-03-19
Update correlation.py

liyao001:
Previously, when people use plotCorrelation.py to generate correlation scatter plot with `--log1p`, the scatter plot is under log1p transformed, but the correlation coefficient is not, which will be misleading. In this PR, I adjusted the order of calculating cor coef, so now the behavior will be consistent.

your pull request :**

 - [ * ] Does the PR contain bugfix?

### DIFF
--- a/deeptools/correlation.py
+++ b/deeptools/correlation.py
@@ -433,7 +433,6 @@ class Correlation:
         """
 
         num_samples = self.matrix.shape[1]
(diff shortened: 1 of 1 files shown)


dpryan79:
I prefer to keep the current implementation, since the correlation shouldn't be affected by how things are plotted. Maybe we can further clarify this in the documentation though so it's clearer?

PR #970, opened 2020-07-10
fix #969

### LINKED ISSUE #969: computeGCBias - matplotlib axis limits error
deeptools 3.4.3
python 3.7.6
matplotlib-base 3.2.2.

command:
computeGCBias -b ~/CLD_demo_project/CNV_visualization_test/03-mapping/cho_horizon_test_merged.bam -g ~/repository/cho_horizon_test/2bit/CHO_Horizon_test.2bit --effectiveGenomeSize 450000 -o cho_horizon_test_merged_gc_freq.tsv --biasPlot cho_horizon_test_merged_gc_freq.png

output:
filter_out: None
extra_sampling_file: None
genome_size: 456350
total_reads: 30552
reads_per_bp: 0.06789333333333333
max_reads: 99.0
min_reads: 0.0
computing frequencies
stepSize: 1
Traceback (most recent call last):
  File "/u/gsemplic/miniconda3/envs/deeptools/bin/computeGCBias", line 12, in <module>
    main(args)
  File "/u/gsemplic/miniconda3/envs/deeptools/lib/python3.7/site-packages/deeptools/computeGCBias.py", line 726, in main
    plotGCbias(args.biasPlot, data, reads_per_gc, args.regionSize, image_format=args.plotFileFormat)
  File "/u/gsemplic/miniconda3/envs/deeptools/lib/python3.7/site-packages/deeptools/computeGCBias.py", line 608, in plotGCbias
    ax1.set_ylim(0 - (y_max * 0.05), y_max * 1.05)
  File "/u/gsemplic/miniconda3/envs/deeptools/lib/python3.7/site-packages/matplotlib/axes/_base.py", line 3577, in set_ylim
    bottom = self._validate_converted_limits(bottom, self.convert_yunits)
  File "/u/gsemplic/miniconda3/envs/deeptools/lib/python3.7/site-packages/matplotlib/axes/_base.py", line 3101, in _validate_converted_limits
    raise ValueError("Axis limits cannot be NaN or Inf")
ValueError: Axis limits cannot be NaN or Inf
(deeptools)

dpryan79:
your pull request :**

### DIFF
--- a/galaxy/wrapper/plotPCA.xml
+++ b/galaxy/wrapper/plotPCA.xml
@@ -78,7 +78,7 @@
             <param name="outFileFormat" value="png" />
             <param name="outFileNameData" value="True" />
             <output name="outFileName" file="plotPCA_result2.png" ftype="png" compare="sim_size" delta="12000" />
-            <output name="output_outFileNameData" file="plotPCA_result2.tabular" ftype="tabular" compare="sim_size" />
+            <output name="output_outFileNameData" file="plotPCA_result2.tabular" ftype="tabular" />
         </test>
     </tests>
     <help>
--- a/galaxy/wrapper/plotPCA.xml
+++ b/galaxy/wrapper/plotPCA.xml
@@ -78,7 +78,7 @@
             <param name="outFileFormat" value="png" />
             <param name="outFileNameData" value="True" />
             <output name="outFileName" file="plotPCA_result2.png" ftype="png" compare="sim_size" delta="12000" />
-            <output name="output_outFileNameData" file="plotPCA_result2.tabular" ftype="tabular" compare="sim_size" delta="12000" />
+            <output name="output_outFileNameData" file="plotPCA_result2.tabular" ftype="tabular" compare="sim_size" />
         </test>
     </tests>
     <help>
--- a/galaxy/wrapper/plotPCA.xml
+++ b/galaxy/wrapper/plotPCA.xml
@@ -78,7 +78,7 @@
             <param name="outFileFormat" value="png" />
             <param name="outFileNameData" value="True" />
             <output name="outFileName" file="plotPCA_result2.png" ftype="png" compare="sim_size" delta="12000" />
-            <output name="output_outFileNameData" file="plotPCA_result2.tabular" ftype="tabular" />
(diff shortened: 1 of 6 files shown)


LeilyR:
I think this error needs to be fixed on planemo. What do you think if we just merge it regardless? The issue is that the generated pc2 is different from the local one, but the local one is correct. At least I get the same value when i run it locally. I can also fix it if you can direct me to where I can find `galaxy-dev/test/functional/test_toolbox.py`

galaxy/wrapper/plotPCA.xml
```diff
@@ -78,7 +78,7 @@
             <param name="outFileFormat" value="png" />
             <param name="outFileNameData" value="True" />
             <output name="outFileName" file="plotPCA_result2.png" ftype="png" compare="sim_size" delta="12000" />
-            <output name="output_outFileNameData" file="plotPCA_result2.tabular" ftype="tabular" />
+            <output name="output_outFileNameData" file="plotPCA_result2.tabular" ftype="tabular" compare="sim_size" delta="12000" />
```
bgruening, line 81: this is not good, why is the difference so high? The idea of such tests is that we spot errors because the output changes dramatically. If the output is actually intended to change than the file `plotPCA_result2.tabular` should be updated to reflect the new "truth".

LeilyR: It is indeed high. plotPCA test in planemo generates 2.49319462166397e-32 for pc2 instead of 1.2325951644078315e-32 in `plotPCA_result2.tabular` . I should not change this though, since it is actually correct. This is actually what the cmd version of deeptools generates too. What is your suggestion?

bgruening: You can check the command that planemo creates. Planemo uses the same version as the CLI version, so either the arguments are different or something strange happens here.

LeilyR: I would say the latter. Maybe some difference in the version of some package or so. 
Though, let me check the planemo args one more time, in case I over looked something.

LeilyR:
@bgruening This is what i was telling you, now all the tests passed even without setting alpha.

LeilyR:
@bgruening That test passed now even without alpha. This is how i meant by being passed time to time.

bgruening:
Wired, we also should move here to container based testing.

Thanks @LeilyR!

PR #999, opened 2020-09-16
return proper numpy `NAN`s in plotFingerprint.py (instead of "NA" strings)

dlaehnemann:
your pull request :**

This problem came up when parsing the OutputQualityMetrics txt output file downstream when it contains the `NA` values in columns. To be exact, the conversion to `float()` in the multiqc module throws an error and leads to the respective plot being skipped. The conversion that fails is this line:
https://github.com/ewels/MultiQC/blob/9760f7561a56c1295a661471d7447abd085b8020/multiqc/modules/deeptools/plotFingerprint.py#L110

According to the [documentation of `float()`](https://docs.python.org/3/library/functions.html#float), this should instead read `nan` to be parsed correctly. And this is what `print(numpy.NAN)` gives (also see [numpy.NAN docs](https://numpy.org/doc/stable/reference/constants.html?highlight=nan#numpy.NaN) and what is used elsewhere in deeptools (a quick grep on the `deeptools/` directory did not give any other results for `"NA"`).


LeilyR:
could you please re-base your PR to the develop branch? Thanks!

dlaehnemann:
@LeilyR , this is rebased onto develop and opened as a new PR onto develop over at PR #1000 . Hope that's what you intended.

PR #1000, opened 2020-09-17
Plot fingerprint: return proper `nan`s

dlaehnemann:
your pull request :**

**Rebased onto develop branch** (compare PR #999 )

This problem came up when parsing the OutputQualityMetrics txt output file downstream when it contains the `NA` values in columns. To be exact, the conversion to `float()` in the multiqc module throws an error and leads to the respective plot being skipped. The conversion that fails is this line:
https://github.com/ewels/MultiQC/blob/9760f7561a56c1295a661471d7447abd085b8020/multiqc/modules/deeptools/plotFingerprint.py#L110

According to the [documentation of `float()`](https://docs.python.org/3/library/functions.html#float), this should instead read `nan` to be parsed correctly. And this is what `print(numpy.NAN)` gives (also see [numpy.NAN docs](https://numpy.org/doc/stable/reference/constants.html?highlight=nan#numpy.NaN) and what is used elsewhere in deeptools (a quick grep on the `deeptools/` directory did not give any other results for `"NA"`).


LeilyR:
thanks! Can you also merge our dev branch to your branch please?

dlaehnemann:
This should be up to date with your `develop` branch, as I rebased onto it. Also, it is `Already up to date` when I do:
```
git checkout develop
git pull origin
git checkout plot-fingerprint_return-proper-NANs
git merge develop
```

So I think this should be fine. I just fixed one of the test failures, as this was simply an expected testing output that needs to read `nan` instead of `NA`, now. But there is one remaining (and seemingly unrelated) error in the deeptools plotPCA tests, where some `Eigenvalue` seems to have changed:
https://dev.azure.com/dpryan79/dpryan79/_build/results?buildId=1791&view=logs&j=23cf4fef-bcb8-5d10-43e4-415f5b313c85&t=4321d142-d313-5ab7-1fea-ce80e4a1ce65&l=1046

No clue, what's happening there, and which value should be correct.

dlaehnemann:
Maybe I'll try another clean start from the develop branch.

dlaehnemann:
Closing in favor of #1001

PR #1017, opened 2020-11-16
fixed xml error brought up  in 1003

LeilyR:
fixed xml error brought up  in #1003


LeilyR:
Hi @dpryan79 Do you know why the galaxy tests failed with this error: 
`ERROR: Your Python version is: 3.7
Galaxy is currently supported on Python 2.7 only.  To run Galaxy,
please download and install a supported version from python.org.  If a
supported version is installed but is not your default, getgalaxy.org
contains instructions on how to force Galaxy to use a different version.`

dpryan79:
@LeilyR Hmm, my guess is that this switched to the 20.09 release of Galaxy and that has a bug in it. Let me check

dpryan79:
OK, I had hard-coded Galaxy version 18.05 into the tests a few years ago since I knew that was working. For some reason that version tag must have just been updated to remove python 3 support, even though it used to be allowed (presumably someone found a show-stopper bug). I'll try updating the tests to use a more recent version of Galaxy and see if that solves the problem.

LeilyR:
thanks a million @dpryan79!

dpryan79:
I'm not sure what the issue is with github actions, but at least azure is running.

LeilyR:
indeed, and now the errors are different and are specific to a single test.

dpryan79:
The plotPCA error is due to changes in matplotlib (or maybe numpy) versions and it's best to just edit the test file so the test passes (basically small non-zero values will change from time to time). The other failure deserves investigation.

LeilyR:
Thanks a lot! I will have a look at them.

Issue #1013, opened 2020-11-12
Plotly causes errors in utilities.py


BJWiley233:
Hi,
When I try to run plotly output with the heatmap I get error in utilities.py `TypeError: 'map' object is not subscriptable` because you are trying to subscript a map which is not allowed in Python 3.  We need to update this if using Python 3 to `C = list(map(np.uint8, np.array(cmap(k * h)[:3]) * 255))`

In addition may want to look at matplotlib deprecation warning at `correlation.py:306`
```

MatplotlibDeprecationWarning:

You are modifying the state of a globally registered colormap. In future versions, you will not be able to modify a registered colormap in-place. To remove this warning, you can make a copy of the colormap first. cmap = copy.copy(mpl.cm.get_cmap("jet"))
```

LeilyR:
it has been taken care of in #1018. already in develop.

PR #1018, opened 2020-11-16
Fix 1013

LeilyR:
inspired by #1013


LeilyR:
Hi @dpryan79,  sorry to bother you. Could you please have a look at [this](https://github.com/deeptools/deepTools/pull/1018/checks?check_run_id=1478936777#step:3:138). I think it needs to be fixed under account. Let me know if there anything I can do for it. Thanks a lot!

dpryan79:
@LeilyR I'll see what I can do

LeilyR:
@dpryan79 Thanks a lot

LeilyR:
Just to let you know, this pc2 value changes frequently between those two numbers, sometime is 1.2325951644078315e-32 some other time is 2.49319462166397e-32. I have seen both of them.

dpryan79:
Yeah, I think it depends on the CPU that gets used.

dpryan79:
@LeilyR can you check computeMatrixOperations locally? It's running rbind and I have no clue why it's immediately throwing an error.

LeilyR:
@dpryan79  I have already tried it locally and it fails there too. I was also really confused, since it says it cannot find the `info` command. with the updated galaxy version also master branch fails for this test, Though I could  get to run the 3.4.3 version locally with no issue. 3.4.3 is the latest version before introducing `dataRange`. What I do not understand is that if that is the problem why rbind test passes. Also `info` used to pass before we change the  galaxy version.  I will have a look at it again, but if you also have any opinion I would be happy to hear. Again thank a million for looking into it.

dpryan79:
I mean on the command line, does it run properly there with `computeMatrixOperations rbind ...`? The data is under galaxy/wrapper/test-data.

LeilyR:
I see ok thanks , i will check

LeilyR:
Hi @dpryan79 , I got to fix the galaxy issues. However there is an issue with the tests running under your account. The error is: 
` Error: The `add-path` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/`
Could you please update this? Do you think I should merge my PR despite those failures?

dpryan79:
The azure tests are all green so it's safe to merge. I'll see if I can get the environment variable set properly. Otherwise I'll just update the nodejs stuff, which should get migrated to the deeptools repo so others can update it more easily.

dpryan79:
It only fails on the annoying plotPCA test. Long-term we might just remove that, since it results only in false positive errors.

LeilyR:
thanks a lot @dpryan79 ! I will merge it then :)

PR #1028, opened 2020-12-10
Add --fixedStep option for bigwigCompare

icebert:
your pull request :**

 - [ Y] Does the PR contain new feature?
 - [ N] Does the PR contain bugfix?
 - [ Y] Does the PR contain documentation changes?
 - [N ] Does the PR contain changes to the galaxy wrapper?


deeptools/bigwigCompare.py
```diff
@@ -93,6 +93,11 @@ def parse_arguments(args=None):
                         'zeros may be wrong and this option should be used ',
                         action='store_true')
 
+    parser.add_argument('--fixedStep',
+                        help='Output in fixed step of bin size (as specified by --binSize), not merging '
```
bgruening, line 97: If anyone has a better phrasing here that would be great. I think the PR is fine but the sentences is a bit confusing for me.

bgruening (approved):
Thanks a lot @icebert!

deeptools/bigwigCompare.py
bgruening: @WardDeb any idea?

WardDeb: Agreed the wording is a bit confusing.
Perhaps:
'Write out all bins (of size --binSize) instead of merging neighbouring bins with equal values.' ?

WardDeb: I'll take it along in develop for now and will make the change afterwards.

PR #1042, opened 2021-02-16
Changing default legend location to 'best'.

blaiseli:
Trying to solve https://github.com/deeptools/deepTools/issues/1041

**Welcome to deepTools GitHub repository! Please check the following regarding
your pull request :**

 - [ ] Does the PR contain new feature?
 - [x] Does the PR contain bugfix?
 - [ ] Does the PR contain documentation changes?
 - [ ] Does the PR contain changes to the galaxy wrapper?

Issue #1041, opened 2021-02-16
legend_location='auto' argument of Profile not accepted by matplotlib


blaiseli:
```
$ deeptools --version
deeptools 3.5.0
$ python --version
Python 3.8.2
$ python -c 'import matplotlib; print(matplotlib.__version__)'
3.3.3
```

The version of matplotlib I use does not allow 'auto' for legend location.

When I run my python script based on deeptools (available here: https://gitlab.pasteur.fr/bli/plotting_scripts/-/blob/master/scripts/create_metagene_profile.py), I get the following error:

```
Traceback (most recent call last):
  File "/home/bli/src/bioinfo_utils/.venv/bin/create_metagene_profile.py", line 396, in <module>
    sys.exit(main())
  File "/home/bli/src/bioinfo_utils/.venv/bin/create_metagene_profile.py", line 384, in main
    compute_matrix(
  File "/home/bli/src/bioinfo_utils/.venv/bin/create_metagene_profile.py", line 285, in compute_matrix
    prof.plot_profile()
  File "/home/bli/src/bioinfo_utils/.venv/lib/python3.8/site-packages/deeptools/plotProfile.py", line 779, in plot_profile
    ax.legend(loc=self.legend_location.replace('-', ' '),
  File "/home/bli/src/bioinfo_utils/.venv/lib/python3.8/site-packages/matplotlib/axes/_axes.py", line 417, in legend
    self.legend_ = mlegend.Legend(self, handles, labels, **kwargs)
  File "/home/bli/src/bioinfo_utils/.venv/lib/python3.8/site-packages/matplotlib/legend.py", line 455, in __init__
    raise ValueError(
ValueError: Unrecognized location 'auto'. Valid locations are
        best
        upper right
        upper left
        lower left
        lower right
        right
        center left
        center right
        lower center
        upper center
        center
```

`prof` is a `deeptools.plotProfile.Profile`.

I suspect something has changes either in deeptools or in matplotlib, because this script used to work.

I would suggest replacing `'auto'` by `'best'` for `legend_location` in the `__init__` of `Profile` in  `deeptools/plotProfile.py`

LeilyR:
You are giht, I think this issue happens if one uses the deeptools api. Can you rebase your PR to develop please?

blaiseli:
I tried to do it. Hoefully I didn't make any mistake.

dpryan79:
Ugh, I hate matplotlib :(

LeilyR:
[...]

PR #1043, opened 2021-02-18
changed auto to best

LeilyR:
your pull request :**

 - [x ] Does the PR contain bugfix? #1041 , #1042 

### DIFF
--- a/deeptools/plotProfile.py
+++ b/deeptools/plotProfile.py
@@ -99,7 +99,7 @@ class Profile(object):
                  plot_type='lines',
                  image_format=None,
                  color_list=None,
-                 legend_location='auto',
+                 legend_location='best',
                  plots_per_row=8,
                  label_rotation=0,
                  dpi=200):


dpryan79:
The galaxy tests are expected to fail then, since some of them will compare the images. Can you see if you can determine when matplotlib added "best" so we can update the matplotlib version restriction?

LeilyR:
sure, i will have a look

LeilyR:
the best option seems to be already availble since version2, this was taken care of for the cli , the default was already set to best, but in the init was auto which i suppose will through an error only if one uses the api.  Planemo test seems to be failed at deeptools_bam_pe_fragmentsize where comparing some tables, some little differences between values (241.20000000000002 instead of 241.2). I thought it could be related to the server the tests are run on it, so I tried to re-run the tests but it did not help. If you do not mind I ginore the error for now and merge this little bug fix to develop desipte the error and we see if it comes up we merging to master again and fix it then.
In general I wonder if we should remove this type of tests from planemo, since it really seems that it is not capable of handelign them. Do you have any idea for that?

dpryan79:
It's a double-edged sword. The benefit to keeping them in is we catch more subtle issues with plotting. So I would vote to keep them in, but wouldn't be terribly disappointed if you said, "Gah! I've had enough of these false-positive failures. Let's remove them!" :)

dpryan79:
Sure, :+1: on merging from my side :)

PR #1137, opened 2022-05-03
Updated dependencies with pip-compile

adRn-s:
The `requirements.txt` file is managed by another [tool](https://pip-tools.readthedocs.io/) that consume new `requirements.in` files. I tried this with python 3.9.x and all tests passed (Except 7, that were already failing).

Btw, I branched from `master`, and I'm starting this PR to merge into `develop` which at the moment is 6 commits ahead of master.


adRn-s:
Ok, I've just checked the error is because of the `sanitizer` elements that validate arguments in the galaxy wrapper. Apparently the `planemo` linter is not aware of these elements (valid according to XML spec [here](https://docs.galaxyproject.org/en/master/dev/schema.html#tool-inputs-param-sanitizer)), because I couldn't find any mention of these in their docs. I will open an issue on planemo, and we'll see...

adRn-s:
Ok, now the error is somewhere else downstream the galaxy tests. According to the [debug log here](https://github.com/deeptools/deepTools/runs/6355319828?check_suite_focus=true#step:4:1291) there's been 2 line changes between the new and the expected output. Then at the bottom I see [the actual diff](https://github.com/deeptools/deepTools/runs/6355319828?check_suite_focus=true#step:4:1356)...

```
--- local_file
+++ history_data
@@ -1,4 +1,4 @@
 #plotPCA --outFileNameData
 Component	bowtie2-test1.bam	bowtie2-test1.bam	Eigenvalue
 1	-0.7071067811865476	-0.7071067811865475	4.0
-2	-0.7071067811865475	0.7071067811865476	2.49319462166397e-32
+2	-0.7071067811865475	0.7071067811865476	1.2325951644078315e-32
```

Could it be that the Eigenvalue change (note it's xx e-32) is due to rounding errors? 

PS: we're testing with galaxy 20.09. I will try updating this to 22.05 later.

docs/content/installation.rst
```diff
@@ -10,15 +10,14 @@ Remember -- deepTools are available for **command line usage** as well as for
 Requirements
 -------------
 
-* Python 2.7 or Python 3.x
```
adRn-s, line 13: Is python 2.7 really supported? @LeilyR

bgruening: We should drop python 2.7 if it's not dropped and also python 3.6.

PR #1186, opened 2023-01-20
Fix galaxy tests

lldelisle:


lldelisle:
@adRn-s I managed to fix the galaxy tests. I also changed the CI so we can access the test report. The only thing I don't like is: my `conda update samtools` but this is due to the fact that the ci-actions belong to dpryan79 so it is not easy to change.

PR #1242, opened 2023-08-21
toml install / requirements version boost

WardDeb:
your pull request :**

 - [x] Does the PR contain documentation changes?

This PR tries to update requirements / python setup to more recent versions (>=3.7).

1. pyproject.toml file 
    - dependencies and metadata are defined in the toml file (setup.py & requirements deprecated)
    - scripts / entrypoints are defined here as well (bin folder deprecated)
    - wheels are built setuptools / toml file now (python >= 3.7)
    - deeptools' version is defined in the toml file as well (no more _version.py, keep importlib as a requirement until we drop python 3.7 support).
    - --version flags have been changed to use importlib in all executables.
    - matplotlib version pin has been removed, some changes in the cm.py for newer version.
2.  CI / github actions
    - nosetests are deprecated and ported to pytests
    - pytests now run on proposed wheel with version 3.7 - 3.11
    - dependency on the custom typescript action (https://github.com/deeptools/github-actions) has been deprecated in favor of micromamba/python/miniconda-setup actions
    - planemo tests and pytests use conda environment(.github/test_and_build.yml), historical 'foo' name is dropped.
    - planemo tests expected pngs are slightly changed (due to mpl updates)
    - expected qual metrics in plotfingerprint changed, not 100% where exactly the rounding difference comes from
    - planemo tests run on galaxy 23.1


WardDeb:
@bgruening Do you mind having a look ? 
I'm not sure if the change to 3.10 will bring some difficulties especially with the galaxy version ?

.github/env.yml
```diff
@@ -0,0 +1,11 @@
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - python = 3.10
```
bgruening, line 5: Why do we pin python 3.10? Could we not use >=3.7 or something?

.github/workflows/planemo.yml
```diff
@@ -1,7 +1,13 @@
 name: Planemo
 on: [push, pull_request]
+
 env:
   GALAXY_BRANCH: release_22.05
```
bgruening, line 5: That release brunch should be bumped. 

```suggestion
  GALAXY_BRANCH: release_23.1
```

.github/workflows/planemo.yml
```diff
@@ -10,12 +16,19 @@ jobs:
...
         chunk: [1, 2, 3]
     steps:
-    - uses: actions/checkout@v1
-    - uses: "dpryan79/github-actions/@master"
+    - uses: actions/checkout@v3
+    - uses: mamba-org/setup-micromamba@main
+      with:
+        environment-file: .github/planemo.yml
+        cache-downloads: true
+        environment-name: foo
```
bgruening, line 24: can we name that differently, that might help in during debugging?

bgruening: I would not use a different yml file for planemo. We need to ensure its the same that we use for testing the unit-tests isn't it? We could just do `pip install planemo` after we installed deeptools maybe? Am I missing something?

.github/workflows/planemo.yml
```diff
@@ -27,7 +40,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7']
+        python-version: ['3.10']
```
bgruening, line 43: we could test multiple versions if we like.
Historically we tested for the lower bound. But we could test for lower and upper-supported python versions.

3.8 and 3.11?

.github/workflows/pypi.yml
```diff
@@ -1,35 +1,33 @@
...
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
-    - name: Setup conda
+    - uses: actions/checkout@v3
+    - uses: mamba-org/setup-micromamba@main
+      with:
+        environment-file: .github/env.yml
+        cache-downloads: true
+        environment-name: foo
```
bgruening, line 21: better name?

.github/workflows/test.yml
```diff
@@ -35,32 +40,77 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
-    - uses: "dpryan79/github-actions/@master"
+    - uses: actions/checkout@v3
+    - uses: mamba-org/setup-micromamba@main
+      with:
+        environment-file: .github/env.yml
+        cache-downloads: true
+        environment-name: foo
```
bgruening, line 48: better name?

.planemo.sh
```diff
@@ -29,6 +29,6 @@ else
 fi
 
 planemo lint ${wrappers}
-planemo test --no_dependency_resolution --galaxy_branch $2 --install_galaxy ${wrappers} 2>&1
+planemo test --no_dependency_resolution --galaxy_python_version 3.10 --galaxy_branch $2 --install_galaxy ${wrappers} 2>&1
```
bgruening, line 32: I don't assume that is necessary. Galaxy will just pick the version that is in the venv.

deeptools/alignmentSieve.py
```diff
@@ -7,15 +7,19 @@
 from deeptools import parserCommon
 from deeptools.bamHandler import openBam
 from deeptools.mapReduce import mapReduce
-from deeptools._version import __version__
+try:  # keep python 3.7 support.
```
bgruening, line 10: another reason why we should keep testing the lowest supported python version imho

bgruening: in addition to the lastest and fastest python version :)

deeptools/test/test_bamCoverage_and_bamCompare.py
```diff
@@ -90,7 +89,7 @@ def test_bam_coverage_skipnas():
         resp = _foo.readlines()
         _foo.close()
         expected = ['3R\t50\t150\t1\n', '3R\t150\t200\t2\n']
-        assert_equal(resp, expected)
+        assert resp == expected, "{} != {}".format(resp, expected)
```
bgruening, line 92: can we use f-strings?

deeptools/test/test_computeMatrixOperations.py
```diff
@@ -20,20 +18,21 @@ def getHeader(fp):
...
 
     def testSubset(self):
         """
         computeMatrixOperations subset
         """
+
         dCorrect = {"verbose": True, "scale": 1, "skip zeros": False, "nan after end": False, "sort using": "mean", "unscaled 5 prime": [0, 0, 0, 0], "body": [1000, 1000, 1000, 1000], "sample_labels": ["SRR648667.forward", "SRR648668.forward", "SRR648669.forward", "SRR648670.forward"], "downstream": [0, 0, 0, 0], "unscaled 3 prime": [0, 0, 0, 0], "group_labels": ["genes"], "bin size": [10, 10, 10, 10], "upstream": [0, 0, 0, 0], "group_boundaries": [0, 196], "sample_boundaries": [0, 100, 200, 300, 400], "max threshold": None, "ref point": [None, None, None, None], "min threshold": None, "sort regions": "no", "proc number": 20, "bin avg type": "mean", "missing data as zero": False}
         oname = "/tmp/subset.mat.gz"
         args = "subset -m {} --sample SRR648667.forward SRR648668.forward SRR648669.forward SRR648670.forward -o {}".format(self.matrix, oname)
+        print(args)
```
bgruening, line 35: should be removed

bgruening (commented):
Cool, a few comments inline :)

Thanks @WardDeb

.github/env.yml
WardDeb: Initially I pinned this since I couldn't get planemo (actually setting up the galaxy instance) working with 3.11, your suggestion to update galaxy branch to 23.1 effectively fixed this: dependency is >=3.7 now

.github/workflows/planemo.yml
WardDeb: env renamed to 'test_and_build' (due to lack of inspiration :) ), and yes, you have a good point:
both planemo / pytests & builds now run from 1 environment.

.github/workflows/planemo.yml
WardDeb: I've included 3.7 & 3.11, though I'm not sure how relevant it is at this stage (planemo_combine_outputs): The tests have already ran at that time, right ?

bgruening (approved):
Very nice. Thanks a lot @WardDeb

PR #1243, opened 2023-08-28
Develop

WardDeb:
-  dependencies and metadata are defined in the toml file (setup.py & requirements deprecated)
-  scripts / entrypoints are defined here as well (bin folder deprecated)
 - wheels are built with setuptools / toml file now (python >= 3.7)
 - deeptools' version is defined in the toml file as well (no more _version.py, keep importlib as a requirement until we drop python 3.7 support).  --version flags have been changed to use importlib in all executables.
 - usage standardized per executable
 - matplotlib version pin has been removed, some changes in the cm.py for newer version.
 - nosetests are deprecated and ported to pytests
 - pytests now run on proposed wheel with version 3.7 - 3.11
 - dependency on the custom typescript action (https://github.com/deeptools/github-actions) has been deprecated in favor of micromamba/python/miniconda-setup actions
 - planemo tests and pytests use conda environment(.github/test_and_build.yml), historical 'foo' name is dropped.
 - planemo tests expected pngs are slightly changed (due to mpl updates) expected qual metrics in plotfingerprint changed, not 100% where exactly the rounding difference comes from
 - planemo tests run on galaxy 23.1
 - stricter label checks and validator in galaxy wrapper
 - deprecation of np types for builtin types


bgruening (approved):
Ship it! Thanks a lot.

PR #1336, opened 2024-11-21
Maturin

WardDeb:
WIP for deepTools 4 overhaul.


bgruening:
@WardDeb would it make sense to split up this PR in many small ones? Maybe even per tool? I guess this makes it easier to review them.

WardDeb:
> @WardDeb would it make sense to split up this PR in many small ones? Maybe even per tool? I guess this makes it easier to review them.

Not so sure to split them up atm (note that for this PR only bamCoverage is properly implemented, only babysteps for bamCompare, so this PR corresponds to one tool..). The number of files changed looks daunting as all the python code got moved into pydeeptools folder, but actually not that much changed.

I'll go ahead and merge this, but keep maturin separated for now. We can discuss moving up to develop next wednesday :)

PR #1345, opened 2024-11-27
Synchronizing Planemo test with Pytests for bamCompare and bamCoverage tool

SaimMomin12:
your pull request :**

---
This PR concerns with adding additional tests to **bamCompare** and **bamCoverage** tool to synchronize them with pytests.


pavanvidem:
@SaimMomin12 did you also check the other way around? including the addtional tests from planemo (if there are any) into pytests?

SaimMomin12:
@pavanvidem yes, the planemo tests are covered in the pytest for these both tools

galaxy/wrapper/bamCompare.xml
```diff
@@ -195,11 +194,91 @@
...
+        </test>
+        <!-- Test with BAM files with skipZeroOverZero -->
+        <test expect_num_outputs="1">
+            <param name="bamFile1" value="testA.bam" ftype="bam"/>
+            <param name="bamFile2" value="testB.bam" ftype="bam"/>
+            <param name="showAdvancedOpt" value="yes"/>
+            <param name="outFileFormat" value="bedgraph"/>
+            <param name="binSize" value="10"/>
+            <param name="type" value="ratio"/>
+            <param name="skipZeroOverZero" value="--skipZeroOverZero"/>
```
wm75, line 257: boolean params (I think this is one) should be tested with `value="true"`/`value="false"` instead of the eventual values.
This is a more stringent test because it checks also whether you typed the correct true/falsevalue in the input section.

SaimMomin12: I see in the `deeptools_macros.xml`, skipZeroOverZero is defined as "select" param type.

wm75: you're right though that seems wrong then by itself.
I'll fix it.

wm75: but maybe in a separate PR :-)

PR #1374, opened 2025-01-23
add plotPCA test

pavanvidem:
@WardDeb `.npz` is in `.gitignore`. Neverthless, I added a required `.npz` test  input file.


bgruening:
I have a general question, would it be good to have those tests added to the current deeptools version as well. So that we can compare results?

pavanvidem:
what do you mean by the current vesion? 3.x?

WardDeb:
> I have a general question, would it be good to have those tests added to the current deeptools version as well. So that we can compare results?

Yea, that's the idea, though only for those functionalities that changed (bamCompare / bamCoverage, computeMatrix & multiBamSummary). Note that this doesn't need to be incoporated in the current version, as for now all the 'old' functionality is still present in 4.0.0 (either through the API, or by bringing back the entrypoints..).

Note that the planemo tests at this point serve as a first checkpoint in this regard, as soon as rewrite is finished (and planemo tests hopefully pass), we can move to the more elaborate setup.

PR #1387, opened 2025-03-18
Viz updates

NavandarM:
Updated graphics


pyproject.toml
```diff
@@ -26,17 +26,18 @@ authors = [
...
-    "plotly >= 4.9",
-    "deeptoolsintervals >= 0.1.8"
+    "numpy >= 2.0",
+    "scipy >= 1.13",
+    "matplotlib >= 3.9",
+    "pysam >= 0.23",
+    "numpydoc >= 1.8",
+    "pyBigWig >= 0.3",
+    "py2bit >= 0.3",
+    "pandas >= 2.2",
```
WardDeb, line 38: is pandas a requirement now ? 
As for the minimum versions, are these correct ? (mainly mpl, scipy, pysam -> how come this changed ?)

WardDeb (approved):

pyproject.toml
caballero: Unfortunately pandas is a dep for scikit-learn, which is added for the PCA analysis. For the others, those are the minimal version compatible with python 3.9, some don't have a real change (pysam, ...)

Issue #1358, opened 2025-01-21
Still numpy version issue...


zyh4482:
Found another conflict about numpy version. computeMatrixOperations.py line 430 has `np.warnings`, which is not an attribute of numpy 2.2. By removing `np.` and add `import warnings`, the rest code executed normally. Not sure if this issue is solved in the develop branch?

PR #1393, opened 2025-04-14
added support for warnings stdlib

AdriTara:
your pull request :**

 - [X] Does the PR contain bugfix?
An error was raising while using `computeMatrixOperations filterValues` on NumPy version 1.26.4. I think NumPy deleted the link to the warnings stdlib as of version 1.21.5 ([See stackoverflow thread](https://stackoverflow.com/questions/76246578/module-numpy-has-no-attribute-warnings)). Therefore, the script computeMatrixOperations.py fails on line 432, raising:

```
np.warnings.filterwarnings('ignore')
AttributeError: module 'numpy' has no attribute 'warnings'. Did you mean: 'hanning'?
```

Therefore, based on the same stackoverflow thread I modified all the files importing numpy to add support for warnings. The other option is to comment the line 432 as suggested in #1358 .

 - [x] Does the PR contain documentation changes?

### DIFF
--- a/deeptools/SES_scaleFactor.py
+++ b/deeptools/SES_scaleFactor.py
@@ -3,6 +3,8 @@
 
 import os
 import numpy as np
+import warnings
+np.warnings = warnings
 
 # own packages
 from deeptools import bamHandler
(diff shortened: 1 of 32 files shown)


WardDeb:
Thanks for the work. Instead of rerouting np warnings through stdlib I'd rather have the specific warnings ignored only in the relevant call context (cfr. #1478).

PR #1431, opened 2026-07-08
feat: lightweight PCA backend (scipy/numpy SVD)

adRn-s:
Replace the sklearn PCA and StandardScaler in Correlation.plot_pca with a scipy/numpy SVD implementation, reproducing sklearn's output (column standardization with population std, deterministic svd_flip sign convention, explained_variance_ from singular values). Drop the now-unused pandas and scikit-learn dependencies.


adRn-s:
## Added PCA test coverage + fixed three inherited `plot_pca` bugs (a04e88c7)

While building a regression test suite for `plotPCA` (to guard the sklearn → scipy/SVD switch), we discovered that the SVD rewrite produces the **same default output as sklearn** — good — but that three pre-existing bugs in the surrounding `plot_pca` logic were carried over unchanged. This commit adds the tests and fixes all three.

### Discoveries

- **The scipy/SVD PCA is numerically equivalent to sklearn** for the default path (matches to `rtol=1e-5`), differing only by per-principal-component sign — expected gauge freedom. Confirmed by running the same suite against both implementations.
- **`--transpose` was fully broken** (in both sklearn and scipy versions): `Wt = np.dot(m, Wt.T)` raises a shape mismatch whenever features ≠ samples, i.e. essentially always. This is the option meant to "match what R does," and it could never run.
- **`--log2` and `--rowCenter` were silent no-ops.** They mutated `self.matrix` *after* `m` had already been copied during `--ntop` filtering, so the transforms never reached the PCA. Output was byte-identical to default.
- **`--ntop` below the sample count crashed** the scatter with an `IndexError` (fewer components than samples to lay out).

### Fixes

1. **Transpose** — dropped the broken `np.dot`; `U * S` already gives each sample's projection onto the PCs, so we just orient it as `(components, samples)`.
2. **log2 / rowCenter** — now applied to a float **copy** of the matrix, **before** variance filtering, so both flags actually affect the result (and `self.matrix` is no longer mutated as a side effect).
3. **Small `--ntop`** — added guards that `sys.exit` with a clear message instead of crashing.

### Tests (`test_plotPCA.py`, 14 passing)

Sign-invariant coordinate + eigenvalue regressions (default and transpose), `--ntop` behavior, variance/eigenvalue consistency, `--PCs` selection, and CLI validation exits. The sign-invariance helper normalizes along the correct component axis (columns for the untransposed table, rows for the transposed one) and stores raw goldens, so the suite is a **fair cross-implementation net** — it passes against both the sklearn and scipy backends rather than only agreeing with itself.

### Note

`pca_back` currently fails to build its Rust extension locally (`maturin` errors), so the PCA tests were run via the working env pointed at this branch's pure-Python source. The rest of the suite (anything needing the compiled `deeptools.hp`) was not exercised here.

WardDeb (approved):
ty, looks good to me !

PR #1445, opened 2026-08-12
Rust backend: parallelism fixes for bigWig writer, alignmentSieve, computeMatrix

adRn-s:
your pull request :**

## Summary

Three bounded parallelism fixes in the Rust backend, found while checking why bamCoverage/bamCompare's bigWig writer, alignmentSieve, and computeMatrix don't scale as well as expected with `-p`. Each commit is independently buildable and testable. Rebased onto current `4.0.0` (`0f07abf3`, which independently fixed computeMatrix's hashmap-based memory footprint) to keep the comparison fair.

1. `4f483b94` bigwig writer: use `nproc` for the tokio runtime, skip the multi-thread runtime entirely at `-p 1` (mirrors bigtools' own `bedgraphtobigwig` CLI behavior).
2. `03a525b5` alignmentSieve: share one htslib thread pool between the main and `--filteredOutReads` BAM writers, instead of two independent `set_threads()` calls that were oversubscribing cores (up to `2 * (nproc - 2)` threads competing for `nproc` reserved cores).
3. `ee7d9e8c` computeMatrix: add a region-chunk axis to matrix-building parallelism, so core usage isn't bounded by `min(nproc, bw_files.len())` on single- or few-sample runs.

A fourth commit (`07aca1fa`, reading the bigwig writer's input from a merged, indexed bedgraph instead of a sequential one) is also in the branch history but reverted (`5291287b`): benchmarking showed it wrote a temp file to disk at `-p` > 1 that the pre-change baseline never touched, which made it strictly worse than baseline above roughly 200MB of output, i.e. most chip/wgs samples. Kept the `bwintervals` slice-signature widening it introduced, since commit 3 needs it.

Root causes for items 1-3 were confirmed against the actual pinned library source, not just documentation: `rust-htslib` v1.0.1 (cloned at that tag) for item 2, and `bigtools` 0.5.8 (cloned at HEAD, one commit past our pinned tag) for item 1.

## Testing

- Full pytest suite passes after every commit (231/232 locally, same one pre-existing unrelated failure as `4.0.0`).
- Manual smoke tests: byte-identical or diff-identical output across `-p 1/4/12` for the affected tools.
- Benchmarked on raven (24-thread SLURM profile, 10 reps per sample, human + triticum references). Summary, full numbers and mechanism writeups are in the PR comments below:
  - bamCoverage/bamCompare/alignmentSieve/multibamSummary: at parity with baseline (~1.0x, ~1.0x mem) everywhere except bamCoverage on low-depth/large-genome samples (triticum chip/rna), where the bigwig writer's per-chromosome parallel encode gives a real 1.5-3.6x speedup, no memory cost. WGS samples see no gain in either species since read-depth-bound coverage computation dominates wall time there, not the write stage.
  - computeMatrix: flat parity (1.0x) when the input already has enough bigwig files to saturate `-p 24` on the file axis alone (15 files here). With few files (3 here), the region-chunking commit still gives a real ~1.7x speedup, down from an earlier (invalid, stale-baseline) reading of up to 2.23x now that both sides share `4.0.0`'s hashmap-to-vector fix. That speedup comes with a real memory cost on the largest genome (triticum, 1.19x baseline RSS), not the near-free win it first looked like.

cc colleagues for input before this merges.

### DIFF
--- a/src/alignmentsieve.rs
+++ b/src/alignmentsieve.rs
@@ -4,6 +4,7 @@ use pyo3::prelude::*;
 use pyo3::types::PyList;
 use rust_htslib::bam::record::CigarString;
 use rust_htslib::bam::{self, Header, Read, Reader, Writer};
+use rust_htslib::tpool::ThreadPool;
 use std::collections::HashMap;
 use std::fs::File;
 use std::io::{BufWriter, Write};
@@ -108,7 +109,13 @@ pub fn r_alignmentsieve(
         None,
     );
 
-    // Open output writers
+    // Open output writers. When both the main and filtered-out BAM writers are active,
+    // give them a single shared htslib thread pool instead of calling `set_threads` on
+    // each independently: `set_threads` spins up its own private pool per call, so two
+    // independent calls with `writerthreads` each would request `2 * writerthreads` BGZF
+    // compression threads total, oversubscribing the `writerthreads` budget we intended.
+    let writer_pool = ThreadPool::new(writerthreads as u32).ok();
+
     let mut obam = if !bed {
         Some(
             Writer::from_path(ofile, &header, bam::Format::Bam)
@@ -118,7 +125,14 @@ pub fn r_alignmentsieve(
         None
     };
     if let Some(ref mut w) = obam {
-        let _ = w.set_threads(writerthreads);
+        match &writer_pool {
+            Some(pool) => {
+                let _ = w.set_thread_pool(pool);
+            }
+            None => {
+                let _ = w.set_threads(writerthreads);
+            }
+        }
     }
     let mut obed = if bed {
         Some(BufWriter::new(File::create(ofile).unwrap_or_else(|e| {
@@ -143,7 +157,14 @@ pub fn r_alignmentsieve(
         None
     };
     if let Some(ref mut w) = ofilterbam {
-        let _ = w.set_threads(writerthreads);
+        match &writer_pool {
+            Some(pool) => {
+                let _ = w.set_thread_pool(pool);
+            }
+            None => {
+                let _ = w.set_threads(writerthreads);
+            }
+        }
     }
     let mut ofilterbed = if write_filters && bed {
         Some(BufWriter::new(
--- a/src/bamcompare.rs
+++ b/src/bamcompare.rs
@@ -395,7 +395,7 @@ pub fn r_bamcompare(
     } else {
         Box::new(raw_lines)
     };
-    write_covfile(lines, ofile, ofiletype, chromsizes);
+    write_covfile(lines, ofile, ofiletype, chromsizes, nproc);
     Ok(())
 }
 
--- a/src/bamcoverage.rs
+++ b/src/bamcoverage.rs
@@ -289,6 +289,6 @@ pub fn r_bamcoverage(
         println!("Writing output to: {}", ofile);
     }
 
-    write_covfile(lines, ofile, ofiletype, chromsizes);
+    write_covfile(lines, ofile, ofiletype, chromsizes, nproc);
     Ok(())
 }
--- a/src/computematrix.rs
+++ b/src/computematrix.rs
@@ -273,28 +273,45 @@ pub fn r_computematrix(
 
     // Discriminate between reference-point and scale-regions mode.
 
-    let matrix: Vec<Vec<f32>> = pool.install(|| {
-        bw_files
+    // bwintervals is single-threaded per bigwig file (BigWigRead::get_interval takes
+    // &mut self, so one reader can't be shared across threads). Parallelizing only
+    // across bw_files.par_iter() therefore leaves most of a reserved node idle whenever
+    // there are fewer tracks than threads, which is the common case. Add a second
+    // parallelism axis by also chunking the region list per file, mirroring the
+    // "open one reader per worker, reuse across a chunk of regions" pattern bam_pileup
+    // already uses for BAM reading. Chunk count in one dimension is intentionally
+    // rounded up so busy still finishes at nproc-ish task count even with 1 bigwig file.
+    let chunks_per_file = (nproc / bw_files.len().max(1)).max(1);
+    let region_chunks = chunk_ranges(regions.len(), chunks_per_file);
+    let tasks: Vec<(usize, usize, usize)> = (0..bw_files.len())
+        .flat_map(|fi| region_chunks.iter().map(move |&(s, e)| (fi, s, e)))
+        .collect();
+
+    // rayon's par_iter().collect() on an indexed source (a Vec, here) is
+    // order-preserving: `results[k]` always corresponds to `tasks[k]`, regardless of
+    // which task finishes first. That lets the merge below stay a plain sequential
+    // walk instead of needing its own bookkeeping to figure out where each chunk goes.
+    let results: Vec<Vec<Vec<f32>>> = pool.install(|| {
+        tasks
             .par_iter()
-            .map(|i| {
+            .map(|&(fi, s, e)| {
                 bwintervals(
-                    &i,
-                    &regions,
-                    &slopregions,
+                    &bw_files[fi],
+                    &regions[s..e],
+                    &slopregions[s..e],
                     &scale_regions,
                     blacklist_index.as_deref(),
                 )
             })
-            .reduce(
-                || vec![vec![]; regions.len()],
-                |mut acc, vec_of_vecs| {
-                    for (i, inner_vec) in vec_of_vecs.into_iter().enumerate() {
-                        acc[i].extend(inner_vec);
-                    }
-                    acc
-                },
-            )
+            .collect()
     });
+
+    let mut matrix: Vec<Vec<f32>> = vec![Vec::new(); regions.len()];
+    for (&(_fi, s, e), chunk_result) in tasks.iter().zip(results.into_iter()) {
+        for (local_ix, region_ix) in (s..e).enumerate() {
+            matrix[region_ix].extend(chunk_result[local_ix].iter().copied());
+        }
+    }
     matrix_dump(
         sortregions,
         sortusing,
@@ -310,6 +327,27 @@ pub fn r_computematrix(
     );
 
     Ok(())
+}
+
+/// Split `[0, n)` into up to `k` contiguous, roughly-equal ranges. Returns a single
+/// `(0, n)` range if `n == 0` or `k <= 1`, so callers never need to special-case
+/// "no chunking" separately from "one chunk".
+fn chunk_ranges(n: usize, k: usize) -> Vec<(usize, usize)> {
+    if n == 0 || k <= 1 {
+        return vec![(0, n)];
+    }
+    let k = k.min(n);
+    let base = n / k;
+    let rem = n % k;
+    let mut ranges = Vec::with_capacity(k);
+    let mut start = 0;
+    for i in 0..k {
+        let len = base + if i < rem { 1 } else { 0 };
+        let end = start + len;
+        ranges.push((start, end));
+        start = end;
+    }
+    ranges
 }
 
 fn slop_region(
--- a/src/filehandler.rs
+++ b/src/filehandler.rs
@@ -31,7 +31,13 @@ pub fn bam_ispaired(bam_ifile: &str) -> bool {
     return false;
 }
 
-pub fn write_covfile<LI>(lines: LI, ofile: &str, filetype: &str, chromsizes: HashMap<String, u32>)
+pub fn write_covfile<LI>(
+    lines: LI,
+    ofile: &str,
+    filetype: &str,
+    chromsizes: HashMap<String, u32>,
+    nproc: usize,
+)
 where
     LI: Iterator<Item = (String, Value)>,
 {
@@ -51,12 +57,26 @@ where
         }
     } else {
         let vals = BedParserStreamingIterator::wrap_infallible_iter(lines, false);
-        let runtime = tokio::runtime::Builder::new_multi_thread()
-            .worker_threads(1)
-            .build()
-            .expect("Unable to create tokio runtime for bw writing.");
-        let writer = BigWigWrite::create_file(ofile, chromsizes)
+        let mut writer = BigWigWrite::create_file(ofile, chromsizes)
             .unwrap_or_else(|e| panic!("Failed to create output bigwig file '{}': {}", ofile, e));
+        // bigtools spawns one async task per chromosome (zoom aggregation + section
+        // encoding) and funnels their output through a single sequential writer task
+        // that owns the actual file handle; the crate's own channels serialize that
+        // hand-off, so raising worker count only widens the per-chromosome encode
+        // stage, it does not add any concurrent writers to `ofile`.
+        // Mirrors bigtools' own bedgraphtobigwig CLI: skip the multi-thread runtime
+        // entirely at nproc == 1, since it would only ever have one consumer.
+        let runtime = if nproc <= 1 {
+            writer.options.channel_size = 0;
+            tokio::runtime::Builder::new_current_thread()
+                .build()
+                .expect("Unable to create tokio runtime for bw writing.")
+        } else {
+            tokio::runtime::Builder::new_multi_thread()
+                .worker_threads(nproc)
+                .build()
+                .expect("Unable to create tokio runtime for bw writing.")
+        };
         let _ = writer.write(vals, runtime);
     }
 }
--- a/src/filehandler.rs
+++ b/src/filehandler.rs
@@ -1,6 +1,8 @@
 use crate::calc::{max_float, mean_float, median_float, min_float, std_float, sum_float};
 use crate::covcalc::{Bin, Gtfparse, Region, Revalue, Scalingregions};
-use bigtools::beddata::BedParserStreamingIterator;
+use bigtools::bed::bedparser::parse_bedgraph;
+use bigtools::bed::indexer::index_chroms;
+use bigtools::beddata::{BedParserParallelStreamingIterator, BedParserStreamingIterator};
 use bigtools::{BigWigRead, BigWigWrite, Value};
 use flate2::Compression;
 use flate2::read::MultiGzDecoder;
@@ -11,7 +13,8 @@ use std::collections::HashMap;
 use std::fs::File;
 use std::io::prelude::*;
 use std::io::{BufReader, BufWriter, Write};
-use std::path::Path;
+use std::path::{Path, PathBuf};
+use tempfile::Builder;
 
 pub fn bam_ispaired(bam_ifile: &str) -> bool {
     let mut bam = Reader::from_path(bam_ifile)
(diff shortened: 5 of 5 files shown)


adRn-s:
# deepTools4 Rust backend: parallelism follow-ups

Scope: four bounded changes identified while investigating why bamCoverage/bamCompare's
bigWig writer, alignmentSieve, and computeMatrix don't scale as well as expected with
thread count. Each item lists the confirmed root cause, the proposed fix, the risk of
introducing races/regressions, and an alternative to fall back on if raven benchmarks
don't show the expected gain. A dedicated memory section closes the doc, since we
already corrected one profile comment this quarter for conflating thread-count-driven
and data-size-driven memory growth, and at least one of these changes (Item 4) turns a
previously data-size-driven cost into a partly thread-count-driven one.

Repo: `deeptools/deepTools`, base branch `4.0.0`. Files referenced are relative to repo root.

---

## Item 1: bamCoverage/bamCompare, drop the text round-trip and adopt bigtools' own parallel reader

**Root cause.** Each parallel region worker in `bam_pileup` (`src/covcalc.rs:196-491`) formats
its coverage values as tab-separated text into a temp bedgraph file
(`writeln!(outbuf, "{}\t{}\t{}\t{}", chrom, s, e, v)`, `covcalc.rs:439`). `write_covfile`'s
caller (`bamcoverage.rs:221-239`, `bamcompare.rs` equivalent) then reopens every one of those
temp files, reads them line by line, splits on tabs, and re-parses the numbers back out
before handing them to `BigWigWrite`. The values were already in memory as numbers; we pay a
format-then-parse cost, plus N small-file opens, on every run regardless of thread count.

Separately, the actual bigWig write path uses `BedParserStreamingIterator`
(`bigtools::beddata`), which reads records strictly sequentially: one `(chrom, Value)` at a
time inside a single `block_on` loop (`beddata.rs:92-178` in bigtools 0.5.8). Raising
`worker_threads` (already done, see prior branch `explore/bigwig-writer-parallel-threads`)
only parallelizes the per-chromosome zoom/encode stage that consumes this iterator; it does
not parallelize the read itself.

bigtools ships a second iterator, `BedParserParallelStreamingIterator`
(`beddata.rs:181-260`), that reads and encodes multiple chromosomes concurrently straight off
a sorted file on disk, using a byte-offset index built by `index_chroms()`. This is not a
theoretical API: it is what bigtools' own `bedgraphtobigwig` CLI uses by default whenever
input isn't stdin and more than one thread is available
(`bigtools/src/utils/cli/bedgraphtobigwig.rs:92-145`, confirmed by cloning
`jackh726/bigtools` directly, HEAD `6109bd5`, one commit past our pinned `v0.5.8` tag).

**Proposed fix.** Two independent sub-changes, worth landing separately so each can be
measured on its own:

- 1a. Replace the per-region text bedgraph temp files with a single binary-encoded temp file
  per `bam_pileup` call (or per region worker, merged once at the end). Removes the
  format/parse round-trip while keeping the existing memory-bounding behavior (values never
  all live in RAM at once for large/binsize=1 runs).
- 1b. Once the values live in one sorted file on disk (which 1a already produces, since
  region workers are processed in chromosome order via `region_divider`), build a
  `chrom_indices` map and switch from `BedParserStreamingIterator` to
  `BedParserParallelStreamingIterator`, mirroring bigtools' own CLI usage. This gets real
  parallel *reading*, not just parallel encoding.

**Race/correctness risk.** Low. Both changes reuse a bigtools API that bigtools' own authors
use as their default, tested, production path for the identical file-to-bigWig job. No new
synchronization code needs to be written on the deepTools side; ordering and correctness are
owned by bigtools' internal channel plumbing (already verified byte-identical output across
worker counts on `explore/bigwig-writer-parallel-threads`).

**Alternative if benchmarks disagree.** bigtools' own `auto` heuristic only turns on the
parallel path above ~200MB input (`bedgraphtobigwig.rs:111`); below that it silently stays
sequential because the indexing overhead isn't worth it. If 1b shows no wall-time win, or a
regression, on typical deepTools output sizes, keep 1a (binary temp file, still a real win)
and gate 1b behind the same kind of file-size (or region-count) threshold rather than always
enabling it. Also worth an A/B against `write_multipass` (bigtools' two-pass writer,
`bigwigwrite.rs:186-`) which the bigtools CLI defaults to for files under that same ~200MB
mark; deepTools currently always calls single-pass `write()`. This cuts the other way from
1b and should not be adopted blind, only if benchmarks show a genuine win for our typical
output sizes.

---

## Item 2: alignmentSieve, stop the dual-writer thread-pool oversubscription

**Root cause.** Confirmed against `rust-bio/rust-htslib` source directly (cloned at the exact
pinned version, tag `v1.0.1`, commit `b576820`). `Writer::set_threads(n)`
(`src/bam/mod.rs:1149`) calls `hts_set_threads`, which creates a **new, private** htslib
thread pool via `hts_tpool_init` for that file handle alone. `alignmentsieve.rs:35-38` derives
`writerthreads = nproc - 2` once, then calls `.set_threads(writerthreads)` independently on
*both* `obam` (main output) and `ofilterbam` (`--filteredOutReads` output) at lines 98 and
112. When `--filteredOutReads` is set, this spins up two separate `writerthreads`-sized BGZF
compression pools: up to `2 * (nproc - 2)` OS threads contending for `nproc` reserved cores,
not `writerthreads`.

rust-htslib's own test suite (`src/bam/mod.rs:2139-2164`) documents the intended pattern for
this exact situation: build one `tpool::ThreadPool::new(n)` and call `.set_thread_pool(&pool)`
on each reader/writer that should share it.

**Proposed fix.** When `write_filters` is true, construct a single
`crate::tpool::ThreadPool::new(writerthreads)` and call `.set_thread_pool(&pool)` on both
`obam` and `ofilterbam` instead of two independent `.set_threads()` calls. When only the main
writer is active, behavior is unchanged (a single `.set_threads()` call is already correct).

**Race/correctness risk.** None expected. `ThreadPool` is `Clone` (wraps an `Rc<RefCell<...>>`
per the rust-htslib source), designed for exactly this multi-handle sharing, and BGZF block
compression across independent output streams has no cross-stream shared mutable state.

**Alternative if benchmarks disagree.** If sharing one pool measurably slows either writer
down under contention (unlikely, but the whole point of this exercise is to check, not
assume), fall back to giving each writer a fixed fraction of `writerthreads` via two
independent `.set_threads(writerthreads / 2)` calls, strictly worse than a shared pool in
theory, but still fixes the oversubscription bug even without adopting the shared-pool API.

---

## Item 3: skip the multi-thread tokio runtime at `-p 1`

**Root cause.** `write_covfile` (`filehandler.rs`) always builds a
`tokio::runtime::Builder::new_multi_thread()` runtime, even when called with a single worker.
bigtools' own `bedgraphtobigwig` CLI special-cases this
(`bedgraphtobigwig.rs:92-100`): at `nthreads == 1` it builds a `new_current_thread()` runtime
and sets `options.channel_size = 0`, avoiding the overhead of spinning up a thread pool and
inter-thread channels that will only ever have one consumer.

**Proposed fix.** Mirror that exact branch in `write_covfile`: `if nproc <= 1 { current_thread
runtime, channel_size = 0 } else { multi_thread runtime, worker_threads(nproc) }`.

**Race/correctness risk.** None. This is the pattern bigtools' own authors use, and it only
changes behavior in the already-degenerate single-thread case.

**Alternative if benchmarks disagree.** None needed; this is a pure removal of unnecessary
setup cost with no plausible downside. If it somehow doesn't measure as a win, it's a no-op
revert, not a design problem to rethink.

---

## Item 4: computeMatrix, add a region-chunk axis to the matrix-building parallelism

**Root cause.** `r_computematrix` (`computematrix.rs:245-258`) parallelizes the
matrix-building step (the dominant cost for typical runs) only across `bw_files`:

```rust
let matrix: Vec<Vec<f32>> = pool.install(|| {
    bw_files.par_iter()
        .map(|i| bwintervals(&i, &regions, &slopregions, &scale_regions))
        .reduce(...)
});
```

`bwintervals` (`filehandler.rs:455-`) then processes *all* regions for one bigWig file in a
strictly sequential `for` loop, reusing one `BigWigRead` reader. This isn't an oversight:
`BigWigRead::get_interval` takes `&mut self` (confirmed in bigtools 0.5.8 source,
`bigwigread.rs:325`), so the reader's internal seek state can't be shared across threads
without one reader instance per thread. But it does mean core utilization is bounded by
`min(nproc, bw_files.len())`, so a common single-sample or few-sample computeMatrix run leaves
most of a reserved node idle for its entire runtime, regardless of `-p`.

**Proposed fix.** Mirror the pattern `bam_pileup` already uses for BAM reading
(`covcalc.rs:207`, "open BAM file once per thread, reuse across all regions"): split `regions`
(and the parallel `slopregions`) into `chunks_per_file = max(1, nproc / bw_files.len())`
contiguous chunks, and parallelize over `(bw_file, chunk)` pairs instead of just `bw_file`.
Each parallel task opens its own `BigWigRead` for its assigned file and processes its chunk of
regions sequentially, exactly as `bwintervals` does today. The only change is that there can
now be more than `bw_files.len()` concurrent tasks. Results merge back into the same
preallocated `Vec<Vec<f32>>`, indexed by the chunk's region-slice offset, instead of via the
current per-file `.reduce()`.

**Race/correctness risk.** Low, same shape as the existing `bam_pileup` parallelism
(independent reader per task, results merged by index, no shared mutable state). The main
implementation risk is bookkeeping: region ordering must be preserved exactly, since
downstream sorting (`matrix_dump`) assumes `matrix[i]` corresponds to `regions[i]`. Needs a
correctness test comparing chunked vs. unchunked output byte-for-byte on a multi-region,
multi-bigwig fixture, the same way we validated Item 1's bigWig output was identical across
worker counts.

**Alternative if benchmarks disagree.** If chunking regions doesn't help (e.g. because
`bwintervals`' per-region cost is small and dominated by BigWigRead per-task open overhead
when chunks are too fine-grained), tune `chunks_per_file` down: cap total tasks at
`nproc` rather than `nproc / bw_files.len() * bw_files.len()`, or make chunk count adaptive on
region count (skip chunking below some region-count threshold, similar to bigtools' own
200MB `auto` heuristic in Item 1). If the gain is real but small relative to the added open-file
overhead, consider reusing one `BigWigRead` per (file, worker-thread) via a thread-local rather
than per-chunk, so long-running rayon worker threads amortize the open cost across chunks
instead of paying it once per chunk.

---

## Memory tradeoff of the combined change set

We deliberately corrected a profile comment earlier this year (`computeMatrix_dt4` mem_mb)
that mischaracterized a data-size-driven memory cost as thread-count-driven. Worth being
equally careful here: two of these four changes plausibly move some memory cost from
"proportional to data" toward "proportional to thread count," which is exactly the kind of
regression that comment-fix was meant to catch early.

- **Item 1 (temp file merge + parallel reader):** 1a is memory-neutral to slightly better
  (same on-disk footprint, less redundant per-file open/close bookkeeping, no new
  in-memory buffering). 1b is the one to watch: `BedParserParallelStreamingIterator`'s
  internal read-ahead queue holds up to 5 chromosomes' parsed values in flight at once
  (`queued_reads.len() < (4 + 1)`, hardcoded in bigtools 0.5.8's `beddata.rs:227`, not exposed
  as a tunable in this version). That's a genuine new memory cost, and it scales with
  per-chromosome data density (i.e. worse at binsize=1 on complex genomes), not thread count
  directly, but it's a step change from the current strictly-one-chromosome-at-a-time
  processing. Worth a dedicated memory benchmark on our worst-case fixture (many chromosomes,
  binsize=1) before/after, the same way we validated the 24-vs-72-thread computeMatrix
  comparison earlier.
- **Item 2 (shared ThreadPool):** memory-reducing. Fewer OS threads spun up overall when
  `--filteredOutReads` is active (one pool instead of two), so smaller total thread-stack
  footprint, not larger.
- **Item 3 (skip multi-thread runtime at `-p 1`):** memory-reducing, marginally. Avoids
  allocating a multi-thread runtime's worker stacks and channels in the one case where they're
  guaranteed to go unused.
- **Item 4 (computeMatrix region chunking):** the one that introduces a
  thread-count-driven memory line item where none existed before. Today, computeMatrix opens
  `bw_files.len()` `BigWigRead` readers total during matrix building. After this change, it can
  open up to `nproc` readers (one per active `(file, chunk)` task) if `bw_files.len() < nproc`.
  Each `BigWigRead` holds a per-instance decompression/read buffer, not the whole file, so per
  the bigtools source this should be a small, bounded overhead (order of a few MB per open
  reader, not something that scales with genome size), but "should be" needs a real number.
  The dominant memory cost (the `Vec<Vec<f32>>` matrix itself, sized by total regions x tracks)
  does not change, since chunking only changes how it's computed, not its final size. Benchmark
  target: peak RSS for a single-bigwig, high-`-p` computeMatrix run, before/after, to put an
  actual number on "up to nproc open readers" instead of leaving it as a plausible-sounding
  estimate.

**Overall expectation going into raven benchmarking.** Items 2 and 3 should show up as pure
wins or no-ops on both time and memory axes; if they don't, something is wrong with the
implementation, not the theory. Items 1 and 4 are the ones where we should genuinely expect
(and look for) a memory/speed tradeoff, and where the "alternative" sections above are the
real fallback plan rather than a formality.

adRn-s:
Four commits, three "confirmed against pinned source" citations, a spec doc with a section literally titled "Memory tradeoff of the combined change set," and not one single em dash anywhere in sight. Either this was written by someone who's spent way too long reading bigtools internals for fun, or it's the most thoroughly laundered AI slop you've ever reviewed. Benchmarks on raven will settle which... 👍🏽  maybe it's both

adRn-s:
Pushed one more commit (349c5947): gates the parallel indexed-read path (item 1b) behind a size threshold on the merged bedgraph.

This was flagged in the spec as the expected fallback if benchmarks disagreed on item 1, and raven's 24-thread run confirmed it does: bamCoverage/bamCompare on WGS-scale samples regressed to 0.78-0.85x wall time under the always-on parallel path, while chip/rna-scale samples sped up up to 3.8x, with memory flat in both cases. Full numbers and rationale are in the commit message.

Set up a proper A/B benchmark Snakefile (4.0.0 pre-branch vs this branch, side by side, two pixi environments so both binaries coexist) on https://github.com/deeptools/deeptools25/tree/dt4-ab-benchmark, `benchmark/dt4_ab.smk`. Will use that to re-verify the threshold fix above and for any further tuning here.

adRn-s:
Status update, replacing the outdated spec above (minimized, kept for history).

Branch is now rebased onto latest `4.0.0` (117f41e5), which picked up two upstream commits: exhaustive testing (#1443) and gz support (#1446). Final 4 commits:

- `0a97c23a` bigwig writer: use nproc for the tokio runtime, skip it entirely at -p 1
- `abf04e41` alignmentSieve: share one htslib thread pool between the two BAM writers
- `126867b4` computeMatrix: add a region-chunk axis to matrix-building parallelism
- `117f41e5` bigwig writer: parallel indexed read at nproc>1, gated by output size

Worth calling out: the rebase surfaced that upstream's gz-support PR had already restructured `write_covfile` to feed bamCoverage/bamCompare's in-memory iterator straight into bigtools (`BedParserStreamingIterator::wrap_infallible_iter`), no temp file, no text round-trip. That's a cleaner version of what the original spec's "Item 1a" was chasing, and it landed independently on `4.0.0` while this branch was in flight. Reconciled by keeping that as the `nproc <= 1` baseline (zero disk I/O, matches upstream exactly) and only spooling to a merged temp file at `nproc > 1`, where it's needed to build bigtools' index for the size-gated parallel reader from the last commit.

232 tests passing (up from 153 pre-rebase, thanks to #1443's exhaustive suite).

Re-running the raven A/B benchmark (4.0.0 pre-branch vs this branch) now that the rebase is in, still draft until those numbers are in.

adRn-s:
Re-ran the A/B benchmark after rebasing on `117f41e5` and found the "size-gated" fix wasn't actually fixing the regression. The gate only chose how to read back a merged temp bedgraph; it still wrote that temp file first at nproc>1, when the pre-branch baseline never touched disk at all. Above the 200MB threshold this made things strictly worse than baseline, and most outputs cross 200MB easily in bedgraph text form, so the regression hit chip/wgs samples across the board, not just wgs as originally reported.

Reverted `117f41e5` (`052bdf5a`), keeping only the unrelated `bwintervals` slice-signature widening it introduced (needed by the computeMatrix chunking commit). Reran bamCoverage/bamCompare, 10 reps, raven 24-thread profile:

| sample | speedup | mem ratio |
|---|---|---|
| bcov_human_chip | 1.11x | 1.00 |
| bcov_human_rna | 1.02x | 0.99 |
| bcov_human_wgs | 0.98x | 0.97 |
| bcov_triticum_chip | 1.54x | 1.05 |
| bcov_triticum_rna | 3.68x | 1.00 |
| bcov_triticum_wgs | 0.98x | 1.01 |
| bcom_human_chip | 1.00x | 0.98 |
| bcom_triticum_chip | 0.99x | 0.96 |

Back at parity with baseline everywhere. The two outliers above 1.5x run identical code to `orig` now, so that's cluster noise, not a real effect.

Still working through computeMatrix/alignmentSieve/multibamSummary numbers before deciding if this is ready to leave draft.

adRn-s:
Quick disclaimer up front: computeMatrix numbers below are stale and a rerun is in progress, see the note at the end before reading too much into them.

Full A/B table after the bigwig-writer revert (10 reps, raven 24-thread profile), for the record:

| tool | sample | speedup | mem ratio |
|---|---|---|---|
| alignmentsieve | human_chip | 1.03x | 1.00 |
| alignmentsieve | human_rna | 1.03x | 1.03 |
| alignmentsieve | human_wgs | 1.04x | 1.00 |
| alignmentsieve | triticum_chip | 1.01x | 0.99 |
| alignmentsieve | triticum_rna | 0.98x | 1.00 |
| alignmentsieve | triticum_wgs | 1.02x | 0.99 |
| bamcompare | human_chip | 1.00x | 0.98 |
| bamcompare | triticum_chip | 0.99x | 0.96 |
| bamcoverage | human_chip | 1.10x | 1.01 |
| bamcoverage | human_rna | 1.02x | 0.99 |
| bamcoverage | human_wgs | 0.98x | 0.97 |
| bamcoverage | triticum_chip | 1.53x | 1.03 |
| bamcoverage | triticum_rna | 3.59x | 1.00 |
| bamcoverage | triticum_wgs | 0.98x | 1.03 |
| computematrix | human_15 | 1.01x | 0.82 |
| computematrix | human_3 | 2.23x | 1.20 |
| computematrix | triticum_15 | 1.03x | 0.85 |
| computematrix | triticum_3 | 1.54x | 1.03 |
| multibamsummary | human_15 | 0.98x | 1.00 |
| multibamsummary | human_3 | 1.00x | 1.00 |
| multibamsummary | triticum_15 | 1.00x | 1.01 |
| multibamsummary | triticum_3 | 0.99x | 1.00 |

alignmentSieve, bamCoverage/bamCompare and multibamSummary are all clean, matching the expected effect of each commit. multibamSummary is untouched by any of this branch's commits and sits flat at ~1.00x, which is a good sanity check that the rest isn't noise.

The computeMatrix row is where it gets complicated. Our `orig` baseline pin (`76ba44b5`) turned out to be stale: it predates a rebase/force-push on `4.0.0` and isn't reachable from the current tip anymore. More importantly, current `4.0.0` picked up `0f07abf3` (drop hashmap for dense vec+indexing in `bwintervals`), a real fix to the same function our region-chunking commit widened the signature of. The two changes aren't redundant: `0f07abf3` changes the per-call memory representation, ours changes how work is split across threads. But the baseline our computeMatrix numbers above were measured against no longer reflects current `4.0.0`, so those speedup/mem numbers can't be trusted as-is.

Rebased the branch onto current `4.0.0` (clean, no conflicts, builds and passes tests locally), repinned both benchmark environments, and a computeMatrix-only rerun against the fair baseline is in progress now. Will post the corrected numbers once that's done.

adRn-s:
Dug into why bamCoverage's triticum_chip (1.53x) and triticum_rna (3.59x) stand out from the rest of the table, which is otherwise close to parity. Short version: it's real, not noise, and it comes from genome size.

`write_covfile` parallelizes the bigwig write itself per chromosome (bigtools spawns one async encode/zoom task per chromosome; our commit switched the tokio runtime from a hardcoded `worker_threads(1)` to `worker_threads(nproc)`, baseline was always serial here regardless of thread count). How much that helps depends on how much per-chromosome work there actually is to split across threads.

Checked the two reference genomes used in the benchmark:

| | total size | contigs | avg chrom size | largest chrom |
|---|---|---|---|---|
| triticum (hexaploid wheat) | 14.5 Gb | 22 | 661 Mb | 831 Mb |
| human | 3.1 Gb | 194 (only 25 >1Mb) | 16 Mb | 249 Mb |

Triticum is about 4.7x bigger genome-wide, in 22 uniformly large chromosomes. Human has more contigs on paper, but 169 of them are unplaced/alt scaffolds that finish almost instantly. The real work is in ~25 chromosomes, the largest of which is a third the size of triticum's largest. So parallelizing the same write stage has far more raw work to divide across threads for triticum.

That also explains why WGS shows no gain in either species (0.98x both) while chip/rna do. WGS read depth is high enough that the upstream BAM coverage computation dominates total wall time, so a faster write stage barely moves the total. Chip/rna have much lower depth, so the write stage (which scales with genome length, not depth) is a bigger share of runtime in both species, and that's where triticum's heavier per-chromosome write cost, now finally parallelized instead of serial, shows up.

computeMatrix numbers are still pending the rerun against the corrected baseline mentioned above, will follow up separately once that's done.

adRn-s:
Updated the PR description to match. Corrected computeMatrix numbers, rerun against the fixed baseline (current `4.0.0` tip, both sides now include the hashmap→vector `bwintervals` fix):

| sample | speedup | mem ratio |
|---|---|---|
| cm_human_15 | 1.00x | 0.91 |
| cm_human_3 | 1.68x | 1.02 |
| cm_triticum_15 | 1.00x | 1.02 |
| cm_triticum_3 | 1.69x | 1.19 |

For reference, the old (stale-baseline) numbers were human_15 1.01x/0.82, human_3 2.23x/1.20, triticum_15 1.03x/0.85, triticum_3 1.54x/1.03.

The `_15` samples (15 bigwig files, enough on their own to saturate 24 threads without any region chunking) now sit at flat parity, which makes sense: the baseline's file-axis parallelism was already using all the threads there, so the region-chunking commit had nothing to add.

The `_3` samples (3 bigwig files, so file-axis parallelism alone leaves most of the 24 threads idle) still show a real ~1.7x speedup. Smaller than the earlier 2.23x, since some of that number really did come from the now-fixed hashmap, but the region-chunking commit is still splitting the 3 files' regions across threads instead of leaving 21 of them idle.

One thing to call out: `cm_triticum_3`'s memory ratio moved the wrong way, 1.19x versus baseline, worse than the old 1.03x reading. Region chunking keeps more region buffers in flight at once, and that cost is more visible now that both sides use the leaner vector-based `bwintervals` instead of a hashmap.

So the chunking commit still earns its complexity on the low-bigwig-count case, but it's a real speed/memory tradeoff rather than a free win. The PR description should say so instead of leading with the old 2.23x number.

